PR #15, opened 2025-03-22
Update security insights loading to ignore case

### LINKED ISSUE #13: Ignore case when searching for security insights file
When running the plugin against a repository with a file at the root named `SECURITY-INSIGHTS.yml` the security insights file cannot be located by the plugin. Proposing the search for the security insights file be case insensitve.

jpower432:
Replace the security insights file name in `loadSecurityInsights` with the found file name from loading the directory contents

Verified with local testing against the `SECURITY-INSIGHTS.yml` in `jpower432/compliance-trestle`. 

Closes #13

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -152,8 +152,9 @@ func (r *RestData) loadSecurityInsights() {
 		return
 	}
 	for _, dirContents := range r.Contents.TopLevel {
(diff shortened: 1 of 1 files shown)


eddie-knight (approved):
Heck yes, brilliant — thank you!

PR #39, opened 2025-03-23
OSPS-BR-01.01

### LINKED ISSUE #41: When sanitizing input, scan will not tokenize variables properly
steps.go/steps_test.go
Package: build_release

Description:

The commented out test TestMultipleVariables fails because the method pullVariablesFromScript does not parse the following line properly, it considers all characters between the first ${{ and last }} as one token.  Ideally this method would create one token per variable so the log output could print the exact token that caused the test to fail.

Example input that causes an improper tokenization.
sed -i 's/{{ TOKEN }}/${{ secrets.TOKEN }}/g' ${{ github.event.review.body }}/.github/pvtr-config.yml

speas038:
Also closes #41


speas038:
Closes #41

jmeridth (approved):
Couple bits but nothing to stop this:
- turn -1 into a constant called not_found and you can remove the comments
- curious why we're using another dep for testing assert function. Testify is usually the industry norm. Is there a feature in this other dep that we use?

Nice work.

eddie-knight:
@jmeridth I believe @speas038 will be unavailable for a little while. Should we merge as-is, or would you like to make your suggested changes first?

jmeridth:
> @jmeridth I believe @speas038 will be unavailable for a little while. Should we merge as-is, or would you like to make your suggested changes first?

Merge as is.

speas038:
Still learning github workflows, didn't realize I added an assert dependency, I think I just used the IDE suggestion for importing an assert function, I'll look into using testify

Issue #24, opened 2025-03-23
Implement OSPS-QA-02


eddie-knight:
The following test requirements need to be implemented:

- [ ] When the package management system supports it, the source code repository MUST contain a dependency list that accounts for the direct language dependencies.
- [ ] When the project has made a release, all compiled released software assets MUST be delivered with a software bill of materials.

vpavankalyan:
I'll be working on this issue today. Could you please assign it to me?

PR #48, opened 2025-03-27
QA-02.01 Test Added #24

vpavankalyan:
Breaking Implementation of QA-02 into 2, the first  QA-02.01 #24 :

- Requirement: When the package management system supports it, the source code [repository](https://baseline.openssf.org/versions/2025-02-25#repository) MUST contain a dependency list that accounts for the direct language dependencies.
- Recommendation: This may take the form a package manager or language dependency file that enumerates all direct dependencies such as package.json, Gemfile, or go.mod.

### DIFF
--- a/evaluation_plans/osps/quality/steps.go
+++ b/evaluation_plans/osps/quality/steps.go
@@ -2,6 +2,7 @@ package quality
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/revanite-io/pvtr-github-repo/evaluation_plans/reusable_steps"
@@ -145,7 +146,7 @@ func requiresNonAuthorApproval(payloadData interface{}, _ map[string]*layer4.Cha
 	if message != "" {
 		return layer4.Unknown, message
 	}
-  	protection := data.Repository.DefaultBranchRef.BranchProtectionRule
+	protection := data.Repository.DefaultBranchRef.BranchProtectionRule
 
 	// Check if reviews are required
 	if !protection.RequiresApprovingReviews {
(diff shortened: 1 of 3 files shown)


evaluation_plans/osps/quality/steps.go
```diff
@@ -188,3 +189,151 @@ func hasOneOrMoreStatusChecks(payloadData interface{}, _ map[string]*layer4.Chan
...
+
+// Known dependency management files for various languages
+var knownManifests = []DependencyManifest{
+	{"go.mod", "go.sum", "Go", true},
+	{"package.json", "package-lock.json", "JavaScript/Node.js", true},
+	{"pom.xml", "", "Java (Maven)", false},
+	{"build.gradle", "gradle.lockfile", "Java (Gradle)", false},
+	{"requirements.txt", "", "Python", false},
+	{"Pipfile", "Pipfile.lock", "Python (Pipenv)", true},
+	{"poetry.lock", "pyproject.toml", "Python (Poetry)", true},
```
trumant, line 229:
```suggestion
	{"pyproject.toml", "poetry.lock", "Python (Poetry)", true},
```

evaluation_plans/osps/quality/steps.go
```diff
@@ -188,3 +189,151 @@ func hasOneOrMoreStatusChecks(payloadData interface{}, _ map[string]*layer4.Chan
...
+	}
+
+	return layer4.Passed, "Dependency management files and SBOM requirements met"
+}
+
+type DependencyManifest struct {
+	Filename    string
+	LockFile    string
+	Language    string
+	IsMandatory bool
```
trumant, line 218: Nit - could this be more clearly named `IsLockFileMandatory`

evaluation_plans/osps/quality/steps.go
```diff
@@ -188,3 +189,151 @@ func hasOneOrMoreStatusChecks(payloadData interface{}, _ map[string]*layer4.Chan
...
+					lockFound := false
+					for _, lockEntry := range entries {
+						if isManifestFile(lockEntry.Name, manifest.LockFile) {
+							result.LockFound = true
+							lockFound = true
+							break
+						}
+					}
+
+					if manifest.IsMandatory && !lockFound {
```
trumant, line 291: https://baseline.openssf.org/versions/2025-02-25#osps-br-05---all-build-and-release-pipelines-must-use-standardized-tooling-where-available-to-ingest-dependencies-at-build-time is the only controls reference to the usage of lockfiles.

I like that you introduced awareness of them in this change, but I wonder if that logic should be moved over the evaluations for the BR control family

eddie-knight: I believe lock files were intentionally omitted here, so I think y'all are on the right track

evaluation_plans/osps/quality/steps.go
vpavankalyan: Removed it as discussed

eddie-knight (approved):

PR #51, opened 2025-03-29
If SI not found, set NeedsReview instead of Failed

eddie-knight:


evaluation_plans/reusable_steps/evaluations.go
```diff
@@ -36,7 +36,7 @@ func HasSecurityInsightsFile(payloadData interface{}, _ map[string]*layer4.Chang
 	}
 
 	if payload.Insights.Header.URL == "" {
-		return layer4.Failed, "Security insights required for this assessment, but file not found"
+		return layer4.NeedsReview, "Security insights required for this assessment, but file not found"
```
trumant, line 39: Should we consider removing this as a discrete step? Potential replacements could be:

- View the presence of insights as a key dependency of the plugin as a whole. Exit execution with error when it's missing
- View the presence of insights as data that contributes to one or potentially of the plugin's assessments. In this scenario, then in the context below:

```go
evaluation.AddAssessment(
	"OSPS-BR-03.01",
	"When the project lists a URI as an official project channel, that URI MUST be exclusively delivered using encrypted channels.",
	[]string{
		"Maturity Level 1",
		"Maturity Level 2",
		"Maturity Level 3",
	},
	[]layer4.AssessmentStep{
		ensureGitHubWebsiteLinkUsesHTTPS,
		reusable_steps.HasSecurityInsightsFile,
		ensureInsightsLinksUseHTTPS,
	},
)
```

there would only be 2 AssessmentSteps:

`ensureGitHubWebsiteLinkUsesHTTPS`: which would return `NeedsReview` when the Insights file is missing
`ensureInsightsLinksUseHTTPS`: which would return `NeedsReview` when the Insights file is missing

The status quo feels like a strange no man's land between those two options to me.

eddie-knight: Hey @trumant, I'm not seeing the strange no man's land you're describing here. Let me know if I'm missing something.

We've gotten feedback from multiple angles (Especially LFX and OpenTelemetry) suggesting that some initial data should be available for users who want to get started with Baseline prior to committing an SI to their repos.

And on the other hand, if we check for the SI in every check where it's needed, we'll have to do something nearly identical to what we have here — except we would be writing the return into all of the corresponding steps rather than putting it at the top as a pre-check.

trumant: ## Overall usage of SI across the implemented control evaluations

See this [Google sheet](https://docs.google.com/spreadsheets/d/1Mm7hDBmUd1aJZ1jnd02zMdx9s-Clf3UNfDXYLX1aS1s/edit?usp=sharing) for the raw data

[19 Level 1 control evaluations](https://docs.google.com/spreadsheets/d/1Mm7hDBmUd1aJZ1jnd02zMdx9s-Clf3UNfDXYLX1aS1s/edit?gid=0#gid=0&fvid=757940695) are currently implemented or a work in progress. 8 of those 19 currently have a dependency on SI.

This PR changes the behavior of 2: BR-03.01 and DO-01.01.

[11 Level 2 control evaluations](https://docs.google.com/spreadsheets/d/1Mm7hDBmUd1aJZ1jnd02zMdx9s-Clf3UNfDXYLX1aS1s/edit?gid=0#gid=0&fvid=80208015) are currently implemented or a work in progress. 5 of those 11 currently have a dependency on SI.

This PR changes the behavior of 1 of those 8: BR-06.01

[5 Level 3 control evaluations](https://docs.google.com/spreadsheets/d/1Mm7hDBmUd1aJZ1jnd02zMdx9s-Clf3UNfDXYLX1aS1s/edit?gid=0#gid=0&fvid=245972690) are currently implemented. 1 of those 5 has a dependency on SI.

This PR makes no changes to level 3 control evaluation.

## More consistent approaches seem advisable

Out of [14 evaluations that rely on SI](https://docs.google.com/spreadsheets/d/1Mm7hDBmUd1aJZ1jnd02zMdx9s-Clf3UNfDXYLX1aS1s/edit?gid=0#gid=0&fvid=453734626), only 3 (those changed here) will return NeedsReview.

What of the remaining 11?

I'd like to see 10 of those remaining 11 all make the same evaluation change you are making here. LE-02.01 does not appear to need any changes for consistency that I can see in its implementation. But, those other 10 could change and be consistent with the spirit of your change here.
 
## More detailed commentary on those 10

#### BR-03.02 - When the project lists a URI as an official distribution channel, that URI MUST be exclusively delivered using encrypted channels.

Current impl:

```go
distributionPoints := data.Insights.Repository.Release.DistributionPoints

if len(distributionPoints) == 0 {
    return layer4.NotApplicable, "No official distribution points found in Security Insights data"
}
```

Given the language of the requirement, users of the project may find numerous "distribution channels" within the various forms of project documentation. Example: those docs include links to the Github releases.

What might incentivize better action by project maintainers:

```go
// no insights file exists
if payload.Insights.Header.URL == "" {
    return layer4.NeedsReview, "Unable to find explicit and machine readable distribution channels. Consider adding a security-insights.yml to explicitly define these for your project."
} else if len(distributionPoints) == 0 {
    return layer4.NeedsReview, "security-insights.yml does not define explicit and machine readable distribution channels. Consider adding URIs in Repository.Release.DistributionPoints."
}....
```

#### DO-02.01 - When the project has made a release, the project documentation MUST include a guide for reporting defects.

Current impl:

```go
if data.Insights.Project.Vulnerability.ReportsAccepted {
    return layer4.Passed, "Repository accepts vulnerability reports"
}
```

Users may find such a guide in the project docs independent of the presence/absence of a SI file. We might incentivize better action by project maintainers by distinguishing between taking a similar approach to the proposal for BR-03.02, i.e. return NeedsReview and vary the message based on SI state

#### DO-03.01 - When the project has made a release, the project documentation MUST contain instructions to verify the integrity and authenticity of the release assets.

Current impl:

```go
if data.Insights.Project.Documentation.SignatureVerification == "" {
	return layer4.Failed, "Signature verification guide was NOT specified in Security Insights data"
}
```

Users may find such instruction in the project docs independent of the presence/absence of a SI file. We might incentivize better action by project maintainers by distinguishing between taking a similar approach to the proposal for BR-03.02, i.e. return NeedsReview and vary the message based on SI state

#### DO-06.01 - When the project has made a release, the project documentation MUST include a description of how the project selects, obtains, and tracks its dependencies.

Current impl:

```go
if data.Insights.Repository.Documentation.DependencyManagement == "" {
	return layer4.Failed, "Dependency management policy was NOT specified in Security Insights data"
}
```

Users may find such a policy in the project docs independent of the presence/absence of a SI file. We might incentivize better action by project maintainers by distinguishing between taking a similar approach to the proposal for BR-03.02, i.e. return NeedsReview and vary the message based on SI state.

#### GV-01.01 - While active, the project documentation MUST include a list of project members with access to sensitive resources.

Perhaps I'm being too literal, but the control is about documentation and yet the current implementation fails if the SI is missing various data elements, none of which are documentation links, all of which are data. We might incentivize better action by project maintainers by distinguishing between taking a similar approach to the proposal for BR-03.02, i.e. return NeedsReview and vary the message based on SI state.

#### GV-01.02 - While active, the project documentation MUST include descriptions of the roles and responsibilities for members of the project.

Again, a case where these very docs may exist, but if the insights file or particular fields do not, we return Failed

#### GV-03.01 - While active, the project documentation MUST include an explanation of the contribution process.

Again, a case where these very docs may exist, but if the insights file or particular fields do not, we return Failed

#### GV-03.02 - While active, the project documentation MUST include a guide for code contributors that includes requirements for acceptable contributions.

Again, a case where these very docs may exist, but if the insights file or particular fields do not, we return Failed

#### QA-04.01 - While active, the project documentation MUST contain a list of any codebases that are considered subprojects or additional repositories.

Again, a case where these very docs may exist, but if the insights file or particular fields do not, we return Failed

#### VM-02.01 - While active, the project documentation MUST contain security contacts.

Again, a case where these very docs may exist, but if the insights file or particular fields do not, we return Failed

eddie-knight: Does this boil down your premise appropriately?

"As a plugin end user, I would like to have a more actionable message when I am told that a check needs manual review due to a missing security insights file"

trumant: That is one area for improvement I'm calling out. I think this somewhat gets tangled up in how the return values from `AssessmentStep`  relate to the ControlEvaluation.Remediation_Guide

The other big thing I'm calling out is consistency of implementation approach and some of my own confusion / a potential broader lack of shared understanding of what the values in https://github.com/revanite-io/sci/blob/main/pkg/layer4/result.go#L9 are fundamentally meant to communicate.

Diving deeper and using `GV-01.02 - While active, the project documentation MUST include descriptions of the roles and responsibilities for members of the project` as an example and thinking about how the AssessmentStep might behave with various repository conditions and plugin implementation conditions

 * **NotRun** or **Unknown** is returned as a simple initial implementation when the control evaluation step is first implemented in the plugin but not yet inspecting either the SI file or even the repository metadata. This communicates to the consumer of the evaluation result that the control exists, but no meaningful evaluation implementation exists.
 * **NeedsReview** is returned when the control evaluation step finds one or more signs of evidence that the control requirement is met, but that evidence is not inconclusive. Some examples of inconclusive evidence for this control include:
   * no security insights file present
   * security insights file present, but no value for the optional `Repository.Documentation.Governance` URL
   * security insights file present, `Repository.Documentation.Governance` does have a URL value, but no examination of the contents of that URL has been done
   * no security insights file present, but file matching `members.md`, `governance.md`, or `maintainers.md` file is found in the root or forge directory
 * **Failed** is returned when the control evaluation step finds conclusive evidence that no `Repository.Documentation.Governance` is available and no file matching `members.md`, `governance.md`, or `maintainers.md` file is found in the root or forge directory
 * **Passed** is extremely rigorous and returned when the control evaluation step has one of:
   * Conclusive evidence via complex automated implementation that roles and responsibilities
   * Some machine-readable attestation that a prior human review was done and found evidence of the control being met - this can possibly exist in the insights file
 * **Not Applicable** is returned when the control evaluation step finds that the repository is archived or it finds a security insight file present with `Repository.Status != "active"`
 
In terms of what can be done to address my confusion, I'm hoping for some discussion here or in our bi-weekly meeting or Slack channel leading to better documentation and schema in https://github.com/revanite-io/sci/blob/main/pkg/layer4/result.go#L9 and in https://github.com/revanite-io/sci/blob/main/schemas/layer-4.cue#L27 which can in turn, hopefully lead to all of the control evaluation steps behaving more consistently with respect to their step results.

trumant: https://github.com/revanite-io/sci/pull/19 has more specific details on how I think we can better deal with multiple different strategies for assessing whether a control requirement is met and the concept of many such assessment strategy implementations having preconditions, like the presence of `security-insights.yml`

trumant (approved):
I'm approving as I think this is an incremental improvement towards the goals I outlined in https://github.com/revanite-io/pvtr-github-repo/pull/51/files#r2025413011

PR #55, opened 2025-04-02
Basic implementation of QA-06.02 and QA-06.03

trumant:
This PR provides a simplistic, but arguably valid implementation for both of these controls, which is to explicitly require human review of the project documentation.

The end-user of the evaluation results  will now get more clear guidance on what to review in the control evaluation result message.


eddie-knight (approved):
Yep... not sure that this one is going to get a scan unless we make an addition to SI.

PR #58, opened 2025-04-06
Fix GraphQL query structure so the test run again

### LINKED ISSUE #57: GraphQL data not returning
When running the plugin, the graphql query fails with a message saying that there are several data elements that don't exist on the corresponding objects.

speas038:
I moved the tree object into the proper hierarchy, and had to remove a few fields that were causing the query to fail.  Seems like those fields were not being used, but need to double check.

Closes #57


speas038:
This pull request closes issue #57

eddie-knight (approved):

PR #62, opened 2025-04-15
feat: adopt goreleaser for managing releases

trumant:
This PR adds goreleaser configuration in preparation for the project cutting official releases in the coming months

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@ config.yml
 
 .vscode
 output
+
+# Go test coverage
+coverage.out
+# Added by goreleaser init:
+dist/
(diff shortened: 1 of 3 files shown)


jmeridth:
I'd prefer to bring over our [release process](https://github.com/privateerproj/privateer/blob/5fff04b31f8f2d88d2836186f417fbd1d54d8f54/.github/workflows/release.yml#L33) from github.com/privateer/privateer.  It is using goreleaser also and does auto versioning and releasing via labels. wdyt?

trumant:
> I'd prefer to bring over our [release process](https://github.com/privateerproj/privateer/blob/5fff04b31f8f2d88d2836186f417fbd1d54d8f54/.github/workflows/release.yml#L33) from github.com/privateer/privateer. It is using goreleaser also and does auto versioning and releasing via labels. wdyt?

Sounds good to me. Love that you have a plan for consistent release approach across multiple related repos.

PR #63, opened 2025-04-15
feat: OSPS-VM-06

### LINKED ISSUE #36: Implement OSPS-VM-06

- [ ] While active, the project documentation MUST include a policy that defines a threshold for remediation of SAST findings.
- [ ] While active, all changes to the project's codebase MUST be automatically evaluated against a documented policy for security weaknesses and blocked in the event of violations except when declared and suppressed as non-exploitable.
- [ ]

speas038:
Closes #36


eddie-knight (approved):

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -25,3 +25,19 @@ func hasSecContact(payloadData interface{}, _ map[string]*layer4.Change) (result
...
+
+
+func sastToolDefined(payloadData interface{}, _ map[string]*layer4.Change) (result layer4.Result, message string) {
+	data, message := reusable_steps.VerifyPayload(payloadData)
+	if message != "" {
+		return layer4.Unknown, message
+	}
+
+	for _,tool := range data.Insights.Repository.Security.Tools {
+		if tool.Type == "SAST" {
```
trumant, line 37: Shame that the enum from https://github.com/ossf/security-insights-spec/blob/main/schema.cue#L116 isn't actually defined in https://github.com/ossf/si-tooling/blob/main/v2/si/security-insights.go

That isn't a critique of your choice here, more an observation of how we can improve

trumant: If for some reason all of the values in https://github.com/ossf/security-insights-spec/blob/main/schema.cue#L122 are `false` for the SAST tool, should this still return `layer4.Passed`?

evaluation_plans/osps/vuln_management/steps_test.go
```diff
@@ -0,0 +1,72 @@
...
+	expectedResult layer4.Result
+	expectedMessage string
+	payloadData interface{}
+	assertionMessage string
+}
+
+
+func TestSastToolDefined(t *testing.T) {
+	
+	testData := []testingData{
```
trumant, line 22: Since https://github.com/ossf/security-insights-spec/blob/main/schema.cue#L114 is optional, it would be great to add a case for no Tools

evaluation_plans/osps/vuln_management/evaluations.go
```diff
@@ -157,7 +157,9 @@ func OSPS_VM_06() (evaluation *layer4.ControlEvaluation) {
 			"Maturity Level 3",
 		},
 		[]layer4.AssessmentStep{
-			reusable_steps.NotImplemented,
+			reusable_steps.HasSecurityInsightsFile,
```
trumant, line 160: The implementation here seems inverted.

https://baseline.openssf.org/versions/2025-02-25#osps-vm-0601 should only be based on `reusable_steps.HasDependencyManagementPolicy`

whereas https://baseline.openssf.org/versions/2025-02-25#osps-vm-0602 should be based on `sastToolDefined`

trumant: Thinking through this a bit more, I'm not sure I see the linkage between the dependency management policy and the `policy that defines a threshold for remediation of SAST findings` language in VM-06.01.

evaluation_plans/osps/vuln_management/steps.go
eddie-knight: oh... that's an interesting edge case, one we wouldn't expect... I'd agree we should fail on such an unexpected state

evaluation_plans/osps/vuln_management/evaluations.go
eddie-knight: Ah... SAST vs SCA. I didn't notice this, and defaulted to my always-on mental thread for supply chain security 😆

speas038: Oh wow, good catch, That is what I meant to put, I'll fix that now

evaluation_plans/osps/vuln_management/steps.go
speas038: I added a check and updated the tests to make sure SAST is explicitly enabled. @trumant @eddie-knight

evaluation_plans/osps/vuln_management/steps_test.go
speas038: Adding that now

evaluation_plans/osps/vuln_management/evaluations.go
speas038: Ok, so leave the dependency-management check out for this one?

PR #69, opened 2025-04-25
feat: implement check for LE-03.02

trumant:
This PR closes https://github.com/revanite-io/pvtr-github-repo/issues/23 by adding logic to check if the project has releases and one or more licenses.

I made the decision to implement the check this way, because of the GitHub docs claim

> GitHub will automatically include links to download a zip file and a tarball containing the contents of the repository at the point of the tag's creation.

in https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ config.yml
 
 .vscode
 output
+
+# go test coverage output
+coverage.out
--- a/evaluation_plans/osps/legal/evaluations.go
+++ b/evaluation_plans/osps/legal/evaluations.go
@@ -88,22 +88,19 @@ func OSPS_LE_03() (evaluation *layer4.ControlEvaluation) {
 		},
 	)
 
-	// Just run the previous assessments for now
 	// TODO: Implement this assessment
-	// evaluation.AddAssessment(
-	// 	"OSPS-LE-03.02",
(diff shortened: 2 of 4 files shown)


eddie-knight:
The reason I skipped this check is because it seems fairly nuanced from my perspective.

I would typically expect to find the license for released software in the code corresponding to that release. But we have end user reports of situations where at release time a license is intentionally modified to an unapproved license, or overruled by a license attached at the top level of the release, alongside things like provenance data.

I say it _seems_ nuanced, because I haven't laid eyes on an example of this happening... But I can imagine that if it happened, we'd need to validate the license by looking at the release assets and the within the released/zipped source code.

trumant:
@eddie-knight I updated the PR comment to provide justification for my decision to favor simplicity over nuance.

I.e. because GitHub will always include the source code in the release, if a project is using GitHub, it seems like we can consider the check implementation I'm proposing as a low to medium quality signal.

I agree that more robust implementation could bear fruit, but my preference is to ship a reasonable confidence check over no check at all.

eddie-knight (approved):
acknowledging that this is on the list of things to improve, I'm happy to see some progress made. Thanks @trumant!

PR #73, opened 2025-04-27
refactor: introduce use of github rest api client

trumant:
This PR refactors the fetch of GitHub Repository and Organization to use the go-github module, rather than reproduce the functioality of that module here.


evaluation_plans/osps/build_release/evaluations.go
```diff
@@ -93,7 +93,6 @@ func OSPS_BR_03() (evaluation *layer4.ControlEvaluation) {
 			"Maturity Level 3",
 		},
 		[]layer4.AssessmentStep{
-			ensureGitHubWebsiteLinkUsesHTTPS,
```
trumant, line 96: This was removed because I can't find a reference to `websiteUrl` in the REST API docs for either the GetRepository or GetOrganization operations.

eddie-knight: Yep, something went sideways there

trumant:
I deliberately chose not to fully integrate the GitHub module until I got some review feedback indicating that the community is supportive of moving in this direction.

I will take that work to bring consistency to the approach as a follow-on item if this is approved for merge.

eddie-knight (approved):
Lots of linting happening, but the core of it is an adjustment to how the data is gathered and the payload is organized.

This makes sense to me. We tried to stay away from the gh module because we observed a big difference in the runtime compared to direct API calls— but I just ran this branch and the main branche several times with the same config, and they appear to be averaging in the same range.

@speas038 Please take a look for awareness as this is an impactful change.

PR #78, opened 2025-04-30
fix: json cannot encode if githubv4.URI is nil

### LINKED ISSUE #76: Plugin crashes with `unexpected EOF` despite valid config
**Description**

Running the plugin results in a crash with unexpected EOF, even with a valid configuration. The plugin process exits with status 2.

**Configuration**
```
loglevel: trace
write-directory: evaluation_results
write: true
output: json
services:
  hashbrown:
    plugin: github-repo

    policy:
      catalogs:
        - OSPS_B
      applicability:
        - Maturity Level 1

    vars:
      owner: rust-lang
      repo: hashbrown
      token: <redacted>
```

**Output**
```
./privateer run
2025/04/29 15:20:10 vars: map[]
2025-04-29T15:20:18.432+0200 [ERROR] tests failed in plugin hashbrown: unexpected EOF
2025-04-29T15:20:18.432+0200 [ERROR] plugin process exited: plugin=/.privateer/bin/github-repo id=447337 error="exit status 2
```

**Expected Behavior**
The plugin should be creating `hashbrown.json` under `evaluation_results/hashbrown`

**Actual Behavior**
Log file (`hashbrown.log`) is created, but json (`hashbrown.json`) is missing

Also attaching  [hashbrown.log](https://github.com/user-attachments/files/19959779/hashbrown.log) file

eddie-knight:
This resolves an edge case where we fail to export JSON results for a repo that doesn't have a contributing guide. If it isn't found, it returned nil, and that nil couldn't be parsed into json.

Changing the type from `githubv4.URI` to `string` solves the bug.

Resolves #76

### DIFF
--- a/data/graphql-data.go
+++ b/data/graphql-data.go
@@ -21,9 +21,9 @@ type GraphqlRepoData struct {
 		Object struct {
 			Tree struct {
 				Entries []struct {
-					Name   string
-					Type   string // "blob" for files, "tree" for directories
-					Path   string
+					Name string
+					Type string // "blob" for files, "tree" for directories
+					Path string
 				}
 			} `graphql:"... on Tree"`
 		} `graphql:"object(expression: \"HEAD:\")"`
@@ -87,7 +87,7 @@ type GraphqlRepoData struct {
 		}
 		ContributingGuidelines struct {
 			Body         string
-			ResourcePath githubv4.URI
+			ResourcePath string
 		}
 		DependencyGraphManifests struct {
 			TotalCount int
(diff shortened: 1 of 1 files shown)


jmeridth (approved):

trumant (commented):
Nice find and fix!

FWIW, we should consider moving this particular data retrieval to the REST API which would likely simplify things a bit. We should also consider introducing a RepositoryContents interface type to define the handful of directory and file contents operations we run against the repo.

Issue #85, opened 2025-05-23
Intermittent 502 Gateway Timeout Errors When Processing Repositories


joanagmaia:
## Summary
We're experiencing intermittent 502 Bad Gateway errors when trying to process certain GitHub repositories — most recently observed with the `perlweeklychallenge-club` repo. This issue is not deterministic: the same repository might succeed on one attempt but fail on another, even with retries.

## Details
**Issue**
The plugin `perlweeklychallenge-club` is failing during payload fetch:

```
non-200 OK status code: 502 Bad Gateway
````

**Error Payload:**

```
{
  "data": null,
  "errors": [
    {
      "message": "Something went wrong while executing your query on 2025-05-09T08:44:58+00:00.
      This may be the result of a timeout, or it could be a GitHub bug.
      Please include `74EA:12EF:1E4F625:1F7700D:681DC07F` when reporting this issue."
    }
  ]
}
````

**Observed Behavior:**
- The issue is intermittent and non-deterministic.
- Reprocessing the same repository sometimes works, sometimes fails.
- In some cases, even after 5 retries, we still fail to retrieve the data.
- This results in slower processing, additional compute overhead, and inconsistent test results.

eddie-knight:
Looks like we need to figure out the field that's too large for the query.

#79 already reduced the query size for repos with large dependency manifests, so now we'll need to see what the next problematic element is is. 

@joanagmaia do you have any examples of repos/targets that are causing this error?

joanagmaia:
> Looks like we need to figure out the field that's too large for the query.
> 
> [#79](https://github.com/revanite-io/pvtr-github-repo/pull/79) already reduced the query size for repos with large dependency manifests, so now we'll need to see what the next problematic element is is.
> 
> [@joanagmaia](https://github.com/joanagmaia) do you have any examples of repos/targets that are causing this error?

Hey 👋 
This one https://github.com/manwar/perlweeklychallenge-club, I'll also ask @epipav to check if we can share more examples besides this one.

eddie-knight:
I have enough confirmation that the error is only coming from the binary check, which retrieves the whole repo tree to check files for binaries. The GraphQL API doesn't support this for large repos, so I'm aiming to refactor that data call to use REST instead.

eddie-knight:
Using REST was still insufficient. The PR that was just merged will resolve the too-large problem by returning `Unknown` if the repo tree call fails.

PR #87, opened 2025-05-31
chore: Refactored rest logic to use more ghClient library

eddie-knight:
This refactor precedes a functional improvement to address #85

This will allow us to loop through and get each repo directory individually, and no longer pull the repo via the GraphQL API (which is failing on large repos).

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -16,22 +16,21 @@ import (
 )
 
 type RestData struct {
-	owner    string
-	repo     string
-	token    string
-	Config   *config.Config
-	Workflow Workflow
-	Insights si.SecurityInsights
-	Releases []ReleaseData
-	Contents Contents
-	Rulesets []Ruleset
-	ghClient *github.Client
+	owner               string
+	repo                string
+	token               string
+	Config              *config.Config
+	WorkflowPermissions WorkflowPermissions
+	Insights            si.SecurityInsights
+	Releases            []ReleaseData
+	Rulesets            []Ruleset
+	contents            RepoContent
+	ghClient            *github.Client
 }
 
-type Contents struct {
-	TopLevel  []*github.RepositoryContent
-	ForgeDir  []*github.RepositoryContent
-	WorkFlows []DirFile
+type RepoContent struct {
+	Content    []*github.RepositoryContent
+	SubContent map[string]RepoContent
 }
 
 type Ruleset struct {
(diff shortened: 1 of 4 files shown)


data/rest-data.go
```diff
@@ -148,46 +132,36 @@ func (r *RestData) checkFile(filename string) (filepath string) {
...
+}
 
-	return err
+func (r *RestData) GetFileContent(path string) (content *github.RepositoryContent, err error) {
+	content, err = r.getSourceFile(r.owner, r.repo, path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve file content for %s: %w", path, err)
+	}
+	if content == nil {
+		return nil, fmt.Errorf("file not found at %s", path)
```
jmeridth, line 162: ❓ should this be `file not found` or `file has no content` or `file empty`? Or that is a length check like you have below?

data/rest-data.go
```diff
@@ -245,32 +219,50 @@ func (r *RestData) loadSecurityInsights() {
...
+				return fmt.Errorf("error getting subdirectory contents for %s: %s", item.GetPath(), err.Error())
+			}
+			c.SubContent[item.GetName()] = RepoContent{
+				Content:    content,
+				SubContent: make(map[string]RepoContent),
+			}
+		}
+	}
+	for path, subContent := range c.SubContent {
+		if err := subContent.getSubDirContents(client, owner, repo); err != nil {
```
jmeridth, line 261: ❓ We're recursively loading all subdirectories and their contents. For very large repositories, this could be slow or resource-intensive. If we expect very large repos, should we consider lazy loading subdirectories or allowing a max recursion depth as a parameter?

jmeridth (approved):
lgtm, a couple non-blocking questions.

data/rest-data.go
eddie-knight: Nil content could be returned if the path was a directory... that may be the only reason a nil would be returned, though I'm not sure. Regarding length, I think the caller should decide whether zero length content equates to an error/failure.

data/rest-data.go
eddie-knight: 🤔 Lemme take a look... Maybe we could only pull subdirectories on request, and then have an exported/public function that can load all content if a test calls for it (ie, binary checking).

It would go against the design principle of up-front data collection, but that's not a strict requirement anyhow.

eddie-knight: Okay — the answer is a definite shmaybe, but it's unravelling a bit as I try to execute it. Making this discussion an issue to follow up on later.

PR #89, opened 2025-06-02
fix: Add error message in the event of error on repo tree query

### LINKED ISSUE #85: Intermittent 502 Gateway Timeout Errors When Processing Repositories

### LINKED ISSUE #88: Recursively loading all subdirectories is uneccessarily slow
❓ We're recursively loading all subdirectories and their contents. For very large repositories, this could be slow or resource-intensive. If we expect very large repos, should we consider lazy loading subdirectories or allowing a max recursion depth as a parameter?

_Originally posted by @jmeridth in https://github.com/revanite-io/pvtr-github-repo/pull/87#discussion_r2118411588_

eddie-knight:
This resolves #85 by adding an `Unknown` status for repos that we cannot query the contents of (502 due to repo size).

This resolves #88 by _not_ changing to recursively get every directory, because this proved to be just as problematic by instantly hitting the rate limit threshold when scanning a large repo.

### DIFF
--- a/evaluation_plans/osps/quality/steps.go
+++ b/evaluation_plans/osps/quality/steps.go
@@ -124,18 +124,24 @@ func statusChecksAreRequiredByBranchProtection(payloadData interface{}, _ map[st
 	return layer4.Passed, "No status checks were run that are not required by branch protection"
 }
 
-// TODO: after 3 layers of depth, make additional API calls if a tree is found.
-// TODO: Examine more than just the file name.
 func noBinariesInRepo(payloadData interface{}, _ map[string]*layer4.Change) (result layer4.Result, message string) {
 	data, message := reusable_steps.VerifyPayload(payloadData)
 	if message != "" {
 		return layer4.Unknown, message
 	}
 
-	if len(data.SuspectedBinaries) == 0 {
-		return layer4.Passed, "No binaries were found in the repository (Note: this check only examines file names at this time)"
+	// TODO: This only checks the top 3 levels of the repository tree
+	// for common binary file extensions and it fails on very large repositories.
+	suspectedBinaries, err := data.GetSuspectedBinaries()
+	if err != nil {
+		data.Config.Logger.Trace(fmt.Sprintf("unexpected response while checking for binaries: %s", err.Error()))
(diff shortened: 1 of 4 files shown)


evaluation_plans/osps/quality/steps.go
```diff
@@ -124,18 +124,24 @@ func statusChecksAreRequiredByBranchProtection(payloadData interface{}, _ map[st
...
 	}
 
-	if len(data.SuspectedBinaries) == 0 {
-		return layer4.Passed, "No binaries were found in the repository (Note: this check only examines file names at this time)"
+	// TODO: This only checks the top 3 levels of the repository tree
+	// for common binary file extensions and it fails on very large repositories.
+	suspectedBinaries, err := data.GetSuspectedBinaries()
+	if err != nil {
+		data.Config.Logger.Trace(fmt.Sprintf("unexpected response while checking for binaries: %s", err.Error()))
+		return layer4.Unknown, "Error while scanning repository for binaries, potentially due to repo size. See logs for details."
```
jmeridth, line 138: I'm a fan of the fail fast method. I'm asking other clarifying questions privately but this is good for me.

jmeridth (approved):

PR #93, opened 2025-06-21
fix: MFA will no longer fail if token is unprivileged

eddie-knight:
This changes to use custom logic instead of `go-githib` to will look for `nil` in the API response to indicate that the token was not authorized to retrieve the MFA data.


speas038:
For future improvements if we ever need to explicitly check the permissions of an access token.   For now the null check works.

GET https://api.github.com/applications/{client_id}/token
Replace {client_id} with your application's client ID.

PR #101, opened 2025-08-06
refactor: replace interface{} with any

zohayb23:
- Replace all `interface{}` with `any` throughout the codebase


eddie-knight (approved):
Tested this locally with `go test` and a run against the privateer repo. Everything seems to be unchanged 👏  LGTM

PR #110, opened 2025-08-13
refactor: make IsCodeRepo testable using GitHub client 

zohayb23:
## Summary
Refactored `IsCodeRepo` to use GitHub client and added tests.

## Changes
- Replace `MakeApiCall()` with `ghClient.Repositories.ListLanguages()`
- Add 3 test cases (success, no languages, error)
- Use mock testing with `mock.GetReposLanguagesByOwnerByRepo`

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -342,22 +342,14 @@ func (r *RestData) GetRulesets(branchName string) []Ruleset {
 }
 
 func (r *RestData) IsCodeRepo() (bool, error) {
-	endpoint := fmt.Sprintf("%s/repos/%s/%s/languages", APIBase, r.owner, r.repo)
-	responseData, err := r.MakeApiCall(endpoint, true)
-	if err != nil {
-		return true, err
-	}
-	languagesUsed := make(map[string]int)
+    languages, _, err := r.ghClient.Repositories.ListLanguages(context.Background(), r.owner, r.repo)
+    if err != nil {
+        return true, err
+    }
 
-	err = json.Unmarshal(responseData, &languagesUsed)
-	if err != nil {
-		return true, err
-	}
+    if len(languages) == 0 {
+        return false, nil
+    }
 
-	r.Config.Logger.Trace(fmt.Sprintf("%v Languages found in repo: %v", len(languagesUsed), languagesUsed))
-	if len(languagesUsed) == 0 {
-		return false, nil
-	}
-
-	return true, nil
+    return true, nil
 }
--- a/data/rest-data_test.go
+++ b/data/rest-data_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/google/go-github/v71/github"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


data/rest-data.go
```diff
@@ -342,22 +342,14 @@ func (r *RestData) GetRulesets(branchName string) []Ruleset {
 }
 
 func (r *RestData) IsCodeRepo() (bool, error) {
-	endpoint := fmt.Sprintf("%s/repos/%s/%s/languages", APIBase, r.owner, r.repo)
-	responseData, err := r.MakeApiCall(endpoint, true)
-	if err != nil {
-		return true, err
-	}
-	languagesUsed := make(map[string]int)
+    languages, _, err := r.ghClient.Repositories.ListLanguages(context.Background(), r.owner, r.repo)
```
trumant, line 345: I know you are relatively new to Go, which is why I offer these improvements that can be made to align your intent with more idiomatic Go style

 - Error handling - Return false, err on error, not true, err. Returning true on error is misleading.
 - Comment - If we really do want to "fail open" so to speak, and identify the repo as a code repo when we failed to get any evidence one way or the other, you should add a comment that explains this departure from typical practice
 - Simplify - You can return the result directly without extra checks.

```
// IsCodeRepo returns true if the repository has any detected programming languages.
func (r *RestData) IsCodeRepo() (bool, error) {
    languages, _, err := r.ghClient.Repositories.ListLanguages(context.Background(), r.owner, r.repo)
    if err != nil {
        return false, err
    }
    return len(languages) > 0, nil
}
```
or, if we want to keep the "fail open" vibes:

```
// IsCodeRepo returns (true, nil) when the repository has any detected programming languages and (true, err) when language analysis fails
func (r *RestData) IsCodeRepo() (bool, error) {
    languages, _, err := r.ghClient.Repositories.ListLanguages(context.Background(), r.owner, r.repo)
    if err != nil {
        return true, err
    }
    return len(languages) > 0, nil
}
```

Also, while I think this is a good first algorithm to fulfill this function's purpose, we might consider going further and considering if the richer language metadata in https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml can be used to further refine our understanding of the repository composition. The additional `type` value there with values of: programming, markup, data, prose, etc

data/rest-data_test.go
```diff
@@ -119,3 +120,60 @@ func TestGetSubdirContentByPath(t *testing.T) {
...
+			},
+			expectedResult: false,
+			expectedError:  false,
+		},
+		{
+			name:           "api error",
+			responses:      []mock.MockBackendOption{},
+			expectedResult: true,
+			expectedError:  true,
+		},
```
trumant, line 157: The [docs on returning errors](https://github.com/migueleliasweb/go-github-mock?tab=readme-ov-file#mocking-errors-from-the-api) suggest this pattern

```suggestion
        {
			name: "api error",
			responses: []mock.MockBackendOption{
				mock.WithRequestMatchHandler(
					mock.GetReposLanguagesByOwnerByRepo,
					http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
						mock.WriteError(w, http.StatusInternalServerError, "github went belly up or something")
					}),
				),
			},

			expectedResult: true,
			expectedError:  true,
		},
```

PR #116, opened 2025-08-26
ci: Update release drafter labels and patterns

eddie-knight:


.github/release-drafter.yml
```diff
@@ -60,7 +60,7 @@ autolabeler:
       - "/^(fix).*/i"
   - label: "infrastructure"
     title:
-      - "/^(infrastructure).*/i"
+      - "/^(infra).*/i"
```
jmeridth, line 63:
```suggestion
      - "/^(infrastructure|infra).*/i"
```

Up for keeping both just in case?

Issue #120, opened 2025-08-28
increase unit test coverage


zohayb23:
# Overview
Conducting a comprehensive review of all test files to identify and implement missing test cases across the codebase. This systematic review ensures our test suite provides thorough coverage of all functionality.

# Missing Test Cases Identified
*(No outstanding items currently identified)*

# Missing Test Cases Completed
- [x] **OSPS-LE-03.02: License verification**
  - Fixed missing "Passed" scenario test case in `TestReleasesLicensed`
  - Added test case for repositories with valid licenses and releases
  - Implemented `stubGraphqlRepo` helper for consistent test data
  - Now covers success path: repositories with releases + valid license → `layer4.Passed`

- [x] **SPDX License Expression Handling**
  - Implemented complex license expression testing
  - Added `splitSpdxExpression` function with proper AND/OR handling
  - Covers complex expressions like "MIT AND Apache-2.0 OR GPL-3.0"
  - Handles edge cases and malformed expressions

- [x] **OSPS-LE-02.01: License Validation (goodLicense function testing)**
  - Fixed `TestGetLicenseList` to test actual production code
  - Eliminated duplicate `testGetLicenseListLogic` function
  - Implemented dependency injection in `getLicenseList` for testability
  - Test now validates the real code path used by `goodLicense` function
  - Added comprehensive unit tests with dependency injection pattern (PR #136)
  - Implemented mock GitHub client for isolated testing
  - Covers all scenarios: valid licenses, invalid licenses, API errors, fallback logic

- [x] **OSPS-VM-01.01: Vulnerability Disclosure Policy Testing (PR #153)**
  - Added `TestHasVulnerabilityDisclosurePolicy` with comprehensive coverage
  - Tests policy present, missing, and invalid payload scenarios
  - Validates Security Insights `SecurityPolicy` field checking
  - Ensures proper error handling and clear messaging

- [x] **OSPS-VM-03.01: Private Vulnerability Reporting Testing (PR #154)**
  - Added `TestHasPrivateVulnerabilityReporting` with 5 comprehensive test cases
  - Tests direct security contact email and security champions fallback
  - Validates `ReportsAccepted` policy enforcement
  - Covers no contact methods and invalid payload scenarios
  - Ensures proper private vulnerability reporting channel validation

jmeridth:
[...]

PR #122, opened 2025-08-29
test: Add success path test case for OSPS-LE-03.02 license verification

zohayb23:
# Description
This PR adds a missing test case for OSPS-LE-03.02 to verify the "Passed" scenario when a repository has both valid releases and a license. 

## Changes Made
- Added new test case "Has releases and license" to `TestReleasesLicensed`
- Added helper function `stubGraphqlRepo` to create test data
- Added `FakeGraphqlRepo` struct to support test data creation
- Added documentation comments explaining test structure and purpose

## Related Issue
Part of #120 - "Increase unit test coverage"


evaluation_plans/osps/legal/steps_test.go
```diff
@@ -8,6 +8,26 @@ import (
...
+// FakeGraphqlRepo provides a minimal implementation of GraphqlRepoData
+// for testing OSPS-LE-03.02 license verification. It only includes
+// the LicenseInfo structure needed for license URL verification.
+type FakeGraphqlRepo struct {
+	Repository struct {
+		LicenseInfo struct {
+			Url string
+		}
+	}
+}
```
eddie-knight, line 20: @zohayb23 How does this compare to the alternative of using the full struct with only the necessary values populated?

evaluation_plans/osps/legal/steps_test.go
```diff
@@ -40,6 +60,22 @@ func TestReleasesLicensed(t *testing.T) {
...
+			name: "Has releases and license",
+			payloadData: data.Payload{
+				RestData: &data.RestData{
+					Releases: []data.ReleaseData{
+						{
+							Name: "v1.0.0",
+						},
+					},
+				},
+				GraphqlRepoData: stubGraphqlRepo("https://api.github.com/licenses/mit"),
```
eddie-knight, line 74: if this is only being run once, is the extra function creating value, compared to just writing it out like you do with RestData above?

zohayb23: The helper function adds value by making the test more readable and maintainable. Instead of having complex GraphQL struct initialization inline, we have a clear function name that tells us exactly what it does.

eddie-knight: Hmm... I'm not sold on it. I think this might be one of the occasions where brevity can increase clarity. Could you switch it to follow the same format as the lines above, using the real object that's being tested directly. If we get to a point where we are doing lots of the same objects, I think this approach will be helpful then.

zohayb23: image
image

The GraphQLRepoData struct requires initializing the entire structure to access specific fields, which led to an IncompatibleAssign error when I tried to mimic the release handling. The screenshots attached illustrate this issue.

Using the FakeGraphqlRepo struct and stubGraphqlRepo function helps keep the test case maintainable and less complex. This approach isolates changes to a single location, reducing redundancy and improving readability. It ensures that the test remains focused on the relevant fields without unnecessary complexity.

I believe this method is the most efficient way to handle the test case, given the constraints of the GraphQLRepoData struct.

eddie-knight (approved):

PR #125, opened 2025-09-04
test: Add unit test for getLicenseList function

zohayb23:
## Overview
This PR adds comprehensive unit tests for the `getLicenseList` function using a function injection pattern to enable mocking of external API calls.

### Problem
`TestGetLicenseList` was calling a duplicate test function instead of the actual `getLicenseList` function, meaning the real production code was never tested.

### Solution
- Modified `getLicenseList` to accept optional `makeApiCall` parameter for dependency injection
- Updated test to call the actual function with mocked API dependency
- Removed duplicate `testGetLicenseListLogic` function
- Added nil-check fallback to use real API in production (`data.MakeApiCall`)

**Files Changed:**
- `steps.go` - Added dependency injection to `getLicenseList`
- `steps_test.go` - Fixed test to call actual function

**Related to #120** - License validation testing improvements


eddie-knight (changes requested):
Been staring at this for a bit, and I can't figure out the goal is here.

There is a function that matches the test name, `getLicenseList` but that function isn't called here. I don't actually see any code being tested... What am I missing?

edit: I see that your comment mentions that `getLicenseList` has hard dependency on the API call, but... if we don't call the function, we don't test the function. What are we actually testing?

Perhaps you can create an adjustment to where `getLicenseList` gets the mock API call in tests and uses the real one at runtime. I'm not personally a fan of maintaining mocks, though, so I don't have more suggestions right now.

zohayb23:
@eddie-knight Thanks for your response. I believe that editing the getLicenseList to where the function gets the mock API call would allow calling the actual function in test cases, which gives us better test confidence and removes code duplication.

jmeridth (approved):

PR #137, opened 2025-09-12
fix: missed some checkouts that are still persisting credentials

jmeridth:


jmeridth:
I'm fixing the failing tests. Seems our example yaml doesn't like limited creds. Good to know.

jmeridth:
Had to run actinolint locally to get the actual error:

`could not parse as YAML: yaml: line 14: found a tab character that violates indentation`

This allowed me to deduce the issue.

eddie-knight (approved):

PR #140, opened 2025-09-21
chore(docs): rm defunct contributor guidance

trumant:


jmeridth:
Why is it defunct now?

trumant:
> Why is it defunct now?

It appears to be copy pasta from another privateer proj guide and the link is broken.

Issue #98, opened 2025-08-04
Loader is being called twice


zohayb23:
I noticed that Loader is being called twice while printing a debug statement. Was this intentional?

<img width="752" height="43" alt="Image" src="https://github.com/user-attachments/assets/efa06beb-cd87-4ff4-938f-8021ef6f92b0" />

PR #148, opened 2025-09-21
fix: duplicate data loading

trumant:
This change closes issue #98 by only passing the loader func to the EvaluationOrchestrator in AddEvaluationSuite

When it was passed twice, it was invoked twice in https://github.com/privateerproj/privateer-sdk/blob/main/pluginkit/evaluation_orchestrator.go#L168

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,8 +36,7 @@ func main() {
 		Version = fmt.Sprintf("%s-%s", Version, VersionPostfix)
 	}
 
-	// NewVessel may take a payload for all suites to reference
-	pvtrVessel := pluginkit.NewEvaluationOrchestrator(PluginName, data.Loader, RequiredVars)
+	pvtrVessel := pluginkit.NewEvaluationOrchestrator(PluginName, nil, RequiredVars)
 
 	requirements, err := baseline.GetAssessmentRequirements()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -36,16 +36,14 @@ func main() {
 		Version = fmt.Sprintf("%s-%s", Version, VersionPostfix)
 	}
 
-	// NewVessel may take a payload for all suites to reference
-	pvtrVessel := pluginkit.NewEvaluationOrchestrator(PluginName, data.Loader, RequiredVars)
+	pvtrVessel := pluginkit.NewEvaluationOrchestrator(PluginName, nil, RequiredVars)
```
trumant, line 39: We could also pass `data.Loader` here and pass `nil` below in `AddEvaluationSuite`, but this approach made more sense to me as the behavior of `NewEvaluationOrchestrator` with respect to its `loader` parameter is undocumented at present.

trumant: @eddie-knight The choice I made here also notably removes the YAML output of the payload data. EvaluationOrchestrator.WriteResults was dumping the full payload along with the evaluation results previously.

I was not a fan of that behavior as it had the side effect of leaking the github credentials because the configuration was also dumped to YAML.

But, I'm calling out that side effect, as I guess its possible that it was viewed as desirable or directly depended upon by someone consuming that YAML output.

PR #158, opened 2025-09-28
feat: improve AC-04.01 evalaution coverage

trumant:
This change adds logic to check the value of `can_approve_pull_request_reviews` as the creation and approval of PRs is arguably a write permission grant.

See https://docs.github.com/en/rest/actions/permissions?apiVersion=2022-11-28#get-default-workflow-permissions-for-a-repository and https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#preventing-github-actions-from-creating-or-approving-pull-requests for reference to behavior of the GitHub configuration and REST API.


data/rest-data.go
```diff
@@ -16,14 +16,15 @@ import (
...
-    Do(req *http.Request) (*http.Response, error)
+	Do(req *http.Request) (*http.Response, error)
 }
 
 type RestData struct {
 	owner               string
 	repo                string
 	token               string
 	Config              *config.Config
+	WorkflowsEnabled    bool
```
trumant, line 27: Whether or not the repository has GitHub Actions enabled is highly relevant to our evaluations as several of them review the contents of `.github/workflows` and use the GitHub REST API to review the configuration of workflows.

Once this change merges, we may also want to update the implementation of OSPS-BR-01.01 as it scans files in `.github/workflows` but if GHA is disabled for the repository it will incorrectly claim that CI input is sanitized because it reviews CI pipelines that are not in use.

eddie-knight (approved):

data/rest-data.go
eddie-knight: Good looking out.

PR #184, opened 2025-10-21
fix: Improve logging when an error is encontered in SI

eddie-knight:
This adds a trace log for cases where an error is encountered while reading the security insights file, as well as a boolean `payload.restdata.insightserror`.

If/when Gemara is updated to halt on `NeedsReview` then this will return the message `An error was encountered while parsing Security Insights content` on every assessment that requires a security insights file.

++ also, added a minor improvement for caching SubDir content instead of caling the gh api every time

Part of [linuxfoundation/insights#746](https://github.com/linuxfoundation/insights/issues/746)

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


eddie-knight:
@zobi I'm gonna be out the next few days, could you push this across the finish line by adding a mock ghclient to `TestCheckFile` similar to how you did in `TestIsCodeRepo`?

trumant (approved):

PR #189, opened 2025-10-28
feat: Add GitHub Actions CI/CD workflow

zohayb23:
Adds GitHub Actions CI/CD workflow with automated build, test, security scanning, and multi-platform Docker image publishing.

Workflow runs automatically on push/PR to `main` or `develop`.


jmeridth:
Integration test, security scan, and lint could all be their own workflows. Build and test another. And then build binaries and push images. You can set dependencies between workflows if needed. This one file is a little too big for my liking.

Looks good otherwise.

jmeridth:
We already have lint and build workflows. That's what I thought. We could use `workflow_run` to fire the CD/security stuff when those 2 pass. 🤔

PR #219, opened 2025-12-21
fix: use GitHub isBinary field for binary detection

### LINKED ISSUE #203: Files mis-identified as binary artifacts
Via LFX Insights, this plugin causes the binary artifacts (OSPS-QA-05.01) check for [guacsec/guac](https://insights.linuxfoundation.org/project/guac/repository/guacsec-guac/security) to fail with:

> Suspected binaries found in the repository: excluded_from_copyright, TECHNICAL_ADVISORY_MEMBERS, Tiltfile, dockerignore

However!

* `excluded_from_copyright` does not exist in the repo (and apparently never has)
* `TECHNICAL_ADVISORY_MEMBERS` is a [plain text file](https://github.com/guacsec/guac/blob/main/TECHNICAL_ADVISORY_MEMBERS) (arguably CSV-shaped)
* `Tiltfile` is a [plain text configuration file](https://github.com/guacsec/guac/blob/main/Tiltfile) used by Tilt
* `dockerignore` is a [plain text configuration file](https://github.com/guacsec/guac/blob/main/dockerignore) used by Docker

pfeifferj:
Replace extension-based binary detection with GitHub's GraphQL isBinary field. The previous implementation flagged extensionless files as binary because the binaryExtensions map included an empty string entry. This caused false positives for common text files like OWNERS, Tiltfile, dockerignore, and TECHNICAL_ADVISORY_MEMBERS.

Fixes #203

### DIFF
--- a/data/graphql-binary-check_test.go
+++ b/data/graphql-binary-check_test.go
@@ -152,25 +152,88 @@ func TestBinaryCheckerIsBinary(t *testing.T) {
 	})
 }
 
-func TestHasNullBytes(t *testing.T) {
+func TestCommonAcceptableFileExtension(t *testing.T) {
 	tests := []struct {
 		name     string
-		content  []byte
+		path     string
 		expected bool
 	}{
-		{"empty content", []byte{}, false},
-		{"text only", []byte("hello world"), false},
-		{"null at start", []byte{0x00, 'a', 'b'}, true},
-		{"null at end", []byte{'a', 'b', 0x00}, true},
-		{"null in middle", []byte{'a', 0x00, 'b'}, true},
-		{"binary content", []byte{0xcf, 0xfa, 0xed, 0xfe, 0x00}, true},
+		{name: "no extension", path: "file", expected: false},
+		{name: "empty extension", path: "file.", expected: false},
+		{name: "funky space extension", path: "file. ", expected: false},
+		{name: "md", path: "file.md", expected: true},
+		{name: "txt", path: "file.txt", expected: true},
+		{name: "yaml", path: "file.yaml", expected: true},
+		{name: "yml", path: "file.yml", expected: true},
+		{name: "json", path: "file.json", expected: true},
+		{name: "toml", path: "file.toml", expected: true},
+		{name: "ini", path: "file.ini", expected: true},
+		{name: "conf", path: "file.conf", expected: true},
+		{name: "env", path: "file.env", expected: true},
+		{name: "sh", path: "file.sh", expected: true},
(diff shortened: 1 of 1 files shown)


pfeifferj:
hi @eddie-knight sorry I only just saw your reply! 

I tested against `gradle/gradle`, and it correctly detected `gradle-wrapper.jar` as binary. 

thank you for sharing your test repo. I took a look at the behaviour: the first difference I noticed was the file sizes which led me down a bit of a rabbit hole, and while I couldn't find an official source on this, I observed that when `isTruncated: true`, the gh graphql api doesn't analyze content and returns `isBinary: null` instead of `true|false`. more digging seems to indicate that blobs are truncated if they exceed 512KB. again, no official docs on this, just trial and error + this old discussion thread [1] seems to vaguely corroborate this.

[1] https://github.com/orgs/community/discussions/23920

data/graphql-binary-check.go
```diff
@@ -112,19 +61,95 @@ type GraphqlRepoTree struct {
...
+}
+
+func (bc *binaryChecker) isBinary(isBinaryPtr *bool, isTruncated bool, path string) bool {
+	if isBinaryPtr != nil {
+		return *isBinaryPtr
+	}
+	if isTruncated {
+		binary, err := bc.checkViaPartialFetch(path)
+		if err != nil {
+			bc.logger.Trace(fmt.Sprintf("failed to check binary status via partial fetch for %s: %s", path, err.Error()))
```
eddie-knight: This makes sense to me.

But it seems like we have three states now: Pass / Fail / Unknown

Gemara assessment steps support returning an unknown state, which becomes a warning with a message instead of a false positive / false negative.

@pfeifferj do you want me to push to your branch to refactor the logic around this to support the three possibilities?

pfeifferj: sure, that's fine for me :)

data/graphql-binary-check.go
```diff
@@ -112,19 +61,95 @@ type GraphqlRepoTree struct {
...
+	if resp.StatusCode != http.StatusPartialContent && resp.StatusCode != http.StatusOK {
+		return false, fmt.Errorf("unexpected status: %d", resp.StatusCode)
+	}
+
+	content, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return false, err
+	}
+
+	return hasNullBytes(content), nil
```
evankanderson: You might want to look at MIME types as one way to separate e.g. `image/png` from `application/*`: https://pkg.go.dev/net/http#DetectContentType

eddie-knight (approved):

PR #147, opened 2025-09-21
feat: add support for branch protections via rules

trumant:
This commit adds support for reading and interpreting the rules applied to the default branch of the repo.

Evaluations that previously only considered the state of branch protection rules will now also consider the state of branch rules.

PR #244, opened 2026-02-21
ci: add OpenTofu IaC and integration test workflow

jmeridth:
Split off from https://github.com/ossf/pvtr-github-repo-scanner/pull/147

## What

Add OpenTofu infrastructure-as-code configuration for managing the example-osps-baseline-level-1 test repository, along with a CI workflow that runs tofu plan on PRs touching the iac/ directory.

## Why

The integration test infrastructure needs to be version-controlled and validated in CI to prevent drift and catch configuration issues before they're applied.

## Notes

- terraform.tfstate is committed intentionally since this is a simple setup without remote state backend
- The CI workflow uses GITHUB_TOKEN which has limited permissions — may need a PAT if tofu plan requires broader access
  - we may introduce Octo-STS here so we don't need a PAT but get more permissions.
  
  
@trumant is the original author. DCO is throwing a fit about me being the signer, only reason Travis is listed as co-author (he is the original author)


jmeridth:
Closing this for now. Will come back

PR #245, opened 2026-02-24
fix: treat empty and whitespace-only URIs as not insecure (OSPS-BR-03.01)

jmeridth:
## What

Updated insecureURI to return false for empty or whitespace-only strings, and added table-driven tests covering empty, whitespace, secure, and insecure URI cases.

## Why

Empty or absent insight links were being flagged as insecure because they don't match any of the allowed protocol prefixes. A missing link is not an insecure link.

[Example error before](https://github.com/privateerproj/.github/actions/runs/22334678700/job/64624416929#step:3:113):

```bash
Error: -24T02:58:46.469Z [ERROR] OSPS-BR-03.01: The following links do not use HTTPS: ,
```

## Notes

- This affects both EnsureInsightsLinksUseHTTPS and DistributionPointsUseHTTPS since they share the insecureURI function.

### DIFF
--- a/evaluation_plans/osps/build_release/steps.go
+++ b/evaluation_plans/osps/build_release/steps.go
@@ -236,7 +236,8 @@ func getLinks(data data.Payload) []string {
 }
 
 func insecureURI(uri string) bool {
-	if strings.HasPrefix(uri, "https://") ||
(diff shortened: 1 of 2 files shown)


jmeridth:
Curious if others think an empty string URI is insecure (i.e., "    ")

Issue #255, opened 2026-03-04
PNG files counted as binaries and OSPS-QA-05.01 fails


tuminoid:
# PNG files counted as binaries and OSPS-QA-05.01 fails

Png files in [metal3-docs](https://github.com/metal3-io/metal3-docs) result in log files from the action:

```
2026-03-02T06:56:14.994Z [ERROR] OSPS-QA-05.01: Suspected binaries found in the repository: component-relationships.png, hwcc_sequence_diagram.png, ironic_authentication_mTLS.png, high-level-arch.png, metal3-banner.pdf, metal3-black.png, metal3-dev-env-transparent.png, metal3-dev-env.png, metal3-website-sticker.png, metal3-white.png, metal3.png, metal3_facet-black-text.png, metal3_facet-whitetext.png
```

They are normal size for pngs and not have executable bits and identify as png:

```
metal3-docs on  main +/- [✓] $ find -name '*.png' -exec ls -l {} \;
-rw-rw-r-- 1 tumi tumi 41709 loka   22  2024 ./docs/presentations/metal3-overview/slide-example-1.png
-rw-rw-r-- 1 tumi tumi 111296 loka   22  2024 ./docs/presentations/metal3-overview/slide-example-2.png
-rw-rw-r-- 1 tumi tumi 46325 loka   22  2024 ./docs/presentations/metal3-overview/metal3-components.png
-rw-rw-r-- 1 tumi tumi 57926 loka   22  2024 ./docs/presentations/metal3-overview/metal3-integration-capi.png
-rw-rw-r-- 1 tumi tumi 163366 loka   22  2024 ./docs/presentations/metal3-overview/slide-example-3.png
-rw-rw-r-- 1 tumi tumi 48500 loka    3 07:38 ./docs/user-guide/src/ironic/images/overview.png
-rw-rw-r-- 1 tumi tumi 5261 loka   22  2024 ./docs/user-guide/src/images/metal3-stack.png
-rw-rw-r-- 1 tumi tumi 16598 loka   22  2024 ./docs/user-guide/src/images/capi-machines.png
-rw-rw-r-- 1 tumi tumi 48652 loka   22  2024 ./docs/user-guide/src/images/metal3-capi-objects.png
-rw-rw-r-- 1 tumi tumi 46688 marras 21  2024 ./docs/user-guide/src/bmo/images/ipa-provisioning.png
[...]

PR #256, opened 2026-03-04
fix: use file mode to distinguish executable binaries from non-executable ones

jmeridth:
Fixes https://github.com/ossf/pvtr-github-repo-scanner/issues/255

## What

Added the Mode field from GitHub's GraphQL TreeEntry to the repo tree struct and updated the binary check to use Unix execute permission bits (mode & 0111) to determine if a binary file is actually executable.

## Why

OSPS-QA-05.01 targets "generated executable artifacts," not all binary files. Non-executable binaries like PNGs and PDFs were being incorrectly flagged, causing false failures for documentation-heavy repositories.

## Notes

- The bitwise check (`mode & 0111`) is used instead of exact mode comparison (`== 33261`) for robustness against non-standard Git implementations, though standard Git only uses 100755 for executables.
- This changes behavior for the `checkViaPartialFetch` fallback path: when `isBinaryPtr` is nil and the file is fetched via HTTP, the mode check is NOT applied. That path lacks the GraphQL mode data, so it retains the existing MIME-based heuristic. Reviewers should consider whether this is acceptable.

### DIFF
--- a/data/graphql-binary-check.go
+++ b/data/graphql-binary-check.go
@@ -23,6 +23,7 @@ type GraphqlRepoTree struct {
 					Name   string
 					Type   string
 					Path   string
+					Mode   int
 					Object *struct {
 						Blob struct {
 							IsBinary    *bool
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
LGTM with question.

PR #305, opened 2026-05-11
chore: bump ospo-reusable-workflows release.yaml to v1.0.1

jmeridth:
## What

Pin the reusable release workflow to v1.0.0 (SHA 592067a69a43d2285f933753d89a7c9d51b96530), remove the local GoReleaser job in favor of the reusable workflow's built-in GoReleaser support, add release.disable to .goreleaser.yaml, and add a Breaking Changes category to release-drafter.

## Why

v1.0.0 of ospo-reusable-workflows broadens the release trigger to include breaking, feature, vuln, and release labels and folds GoReleaser, container image build, attestation, and discussion creation into the reusable workflow itself. Centralizing the GoReleaser step removes duplicated CI definition and keeps version pins on a single hardened source.

## Notes

- The reusable workflow validates that the GoReleaser config sets `release.disable: true` to prevent the goreleaser step from racing the draft release.
- The outer label-filter `if:` block is removed because the v1.0 reusable workflow handles label filtering internally.
- Trigger updated to `pull_request_target` so the workflow can push tags via `GITHUB_TOKEN`.
- Attestation is now generated by the reusable workflow (via `create-attestation: true`).
- The separate `release_image` job was also removed; v1.0.0 of the reusable workflow handles container image build/publish internally.

## Testing

- [ ] Workflow parses successfully on PR (lint via GitHub Actions parser)
- [ ] Dry-run release path validates `release.disable: true` in `.goreleaser.yaml`
- [ ] Verify next merge to `main` triggers the reusable workflow as expected and produces a draft release with attestations

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -8,6 +8,9 @@ template: |
   See details of [all code changes](https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION) since previous release
 
 categories:
+  - title: "💥 Breaking Changes"
+    labels:
+      - "breaking"
   - title: "🚀 Features"
     labels:
       - "feature"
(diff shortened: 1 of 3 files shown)


jmeridth:
Closing. Making an upstream change and will include what was left in this PR after #317 merged.

PR #316, opened 2026-05-13
fix: protect supported catalog IDs in CI

vinayada1:
## Summary

This change makes supported catalog IDs an explicit compatibility contract and verifies that contract in CI.

Older config values should not stop working silently when catalog IDs change. This update protects against that by keeping the supported IDs in one place, checking that those IDs still exist in the bundled catalog data, and exercising the integration path for each supported catalog ID during CI.

## Why

A recent regression showed that catalog ID changes can break previously supported configs without CI catching it. This PR adds a lightweight contract check so compatibility regressions are detected earlier.

## Related issues

Refs (https://github.com/ossf/pvtr-github-repo-scanner/pull/307)


jmeridth:
nit: new lines at ends of files. Note red circles in changed file view. It's a checkbox in vscode to add new lines n save.

eddie-knight (approved):

PR #356, opened 2026-07-03
fix(OSPS-BR-07.01): consult Security Insights when security_and_analysis is null

### LINKED ISSUE #354: OSPS-BR-07.01 false-fails when GitHub security_and_analysis is unavailable (unprivileged scans)
## Summary

`OSPS-BR-07.01` ("Secret scanning is not enabled") false-fails for repositories
that enable GitHub Secret Scanning via Terraform and declare `type: secret-scanning`
in `.github/security-insights.yml`. Reported via
[linuxfoundation/insights#1971](https://github.com/linuxfoundation/insights/issues/1971#issuecomment-4843120944).
Credit: @Kielek for the detailed root-cause analysis.

## Affected repositories (examples)

- `open-telemetry/opentelemetry-dotnet`
- `open-telemetry/opentelemetry-dotnet-contrib`
- `open-telemetry/opentelemetry-dotnet-instrumentation`

All three enable Secret Scanning + Push Protection via the
[OpenTelemetry admin Terraform module](https://github.com/open-telemetry/admin/blob/d8bdd2b60aafb43a9523990046fcadb7ae86c3dc/modules/repository/main.tf#L24-L27)
and declare `type: secret-scanning` in `.github/security-insights.yml`.

## Root cause — early return in `buildSecurityPosture`

In `data/security-posture.go` (lines 24–31), when the GitHub API returns
`security_and_analysis: null`, the function returns an empty struct **before**
reaching the `insightsClaimsSecretsTooling(rd.Insights)` check:

```go
securityConfig := repository.GetSecurityAndAnalysis()
if securityConfig == nil {
    return &RepoSecurityPosture{
        restData: rd,
    }, nil          // ← Security Insights fallback is never evaluated
}
```

GitHub only returns `security_and_analysis` to repository administrators,
organisation owners, or security managers
([REST API docs](https://docs.github.com/en/rest/repos/repos#get-a-repository)).
Scanners running without privileged access always receive `nil` here, so the
`type: secret-scanning` declaration in Security Insights is silently ignored.

## Secondary bug — push protection inferred from wrong field

Lines 32–41 derive both properties from `secret_scanning.status`:

```go
preventsSecretPushing: secretsScanningStatus == "enabled" || insightsClaimsSecretsTooling,
scansForSecrets:       secretsScanningStatus == "enabled" || insightsClaimsSecretsTooling,
```

`secret_scanning_push_protection.status` is never read. When privileged
metadata is available, ordinary secret scanning is treated as proof that push
protection is also on.

## Proposed fix

1. Evaluate `insightsClaimsSecretsTooling(rd.Insights)` **before** the
   `securityConfig == nil` return, and populate `preventsSecretPushing` /
   `scansForSecrets` from it in that branch.
2. When privileged metadata is present, read
   `secret_scanning_push_protection.status` for `PreventsPushingSecrets()`.
3. Add regression tests in `data/security-posture_test.go` covering:
   - `SecurityAndAnalysis == nil` + Security Insights `type: secret-scanning`
   - `secret_scanning.status == "enabled"` with push protection disabled

## Impact

Any unprivileged scanner (including LFX Insights) will false-fail
`OSPS-BR-07.01` for every repository that relies on Security Insights
declarations to satisfy the control.

joanagmaia:
## Summary

Fixes #354

Two bugs caused `OSPS-BR-07.01` to false-fail for unprivileged scans (e.g. LFX Insights) on repos that enable Secret Scanning via Terraform and declare `type: secret-scanning` in `.github/security-insights.yml`:

- **Early return skipped Security Insights fallback.** `buildSecurityPosture` returned an empty struct when `security_and_analysis` was `null` (GitHub only returns this field to admins/org owners/security managers), so the `insightsClaimsSecretsTooling` check was never reached for unprivileged tokens.
- **Push protection inferred from wrong field.** Both `preventsSecretPushing` and `scansForSecrets` derived from `secret_scanning.status`; `secret_scanning_push_protection.status` was never read.

## Changes

- `data/security-posture.go` — evaluate `insightsClaimsSecretsTooling` before the `nil` guard and populate both fields in the nil branch; read `SecretScanningPushProtection.GetStatus()` for `preventsSecretPushing`; guard against `nil` `insights.Repository` in `insightsClaimsSecretsTooling`.
- `data/security-posture_test.go` — update existing enabled-status test; add regression tests for both bugs.

## Test plan

- [ ] `go test ./data/... -v -run TestBuildSecurityPosture` — all cases pass
- [ ] `go test ./...` — full suite green
- [ ] End-to-end: run against `open-telemetry/opentelemetry-dotnet` with an unprivileged token → `OSPS-BR-07.01` now `Passed`
- [ ] Regression: privileged token on repo with scanning on / push protection off → `OSPS-BR-07.01` now `Failed` (partially enabled) instead of wrongly `Passed`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/data/security-posture.go
+++ b/data/security-posture.go
@@ -20,17 +20,20 @@ type RepoSecurityPosture struct {
 }
 
 func buildSecurityPosture(repository *github.Repository, rd RestData) (SecurityPosture, error) {
+	insightsClaimsSecretsTooling := insightsClaimsSecretsTooling(rd.Insights)
 	securityConfig := repository.GetSecurityAndAnalysis()
 	if securityConfig == nil {
 		return &RepoSecurityPosture{
-			restData: rd,
+			restData:              rd,
+			preventsSecretPushing: insightsClaimsSecretsTooling,
+			scansForSecrets:       insightsClaimsSecretsTooling,
 		}, nil
 	}
 	secretsScanningStatus := securityConfig.GetSecretScanning().GetStatus()
-	insightsClaimsSecretsTooling := insightsClaimsSecretsTooling(rd.Insights)
+	pushProtectionStatus := securityConfig.GetSecretScanningPushProtection().GetStatus()
 	return &RepoSecurityPosture{
 		restData:              rd,
-		preventsSecretPushing: secretsScanningStatus == "enabled" || insightsClaimsSecretsTooling,
+		preventsSecretPushing: pushProtectionStatus == "enabled" || insightsClaimsSecretsTooling,
 		scansForSecrets:       secretsScanningStatus == "enabled" || insightsClaimsSecretsTooling,
 		// TODO: consider if SecurityInsights should have a policy doc field in ProjectDocumentation to handle this
 		// definesPolicyForHandlingSecrets: rd.SecurityInsights != nil && ....
@@ -38,7 +41,7 @@ func buildSecurityPosture(repository *github.Repository, rd RestData) (SecurityP
 }
 
 func insightsClaimsSecretsTooling(insights si.SecurityInsights) bool {
-	if insights.Repository.SecurityPosture.Tools == nil {
+	if insights.Repository == nil || insights.Repository.SecurityPosture.Tools == nil {
 		return false
 	}
 	for _, tool := range insights.Repository.SecurityPosture.Tools {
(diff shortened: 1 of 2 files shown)


eddie-knight (approved):
Thanks for getting to the bottom of this @joanagmaia!

PR #346, opened 2026-06-12
feat(quality): add AI-assisted OSPS-QA-06.02 evaluation

vinayada1:
## Summary

Implements an AI-assisted evaluator for **OSPS-QA-06.02**, replacing the hard-coded manual-review placeholder for checking whether contributor documentation explains when and how tests are run.

## Changes

- Maps `OSPS-QA-06.02` to `quality.TestExecutionDocumentation`.
- Sends a conservative evidence boundary (README and CONTRIBUTING content only) to the configured AI provider.
- Records the structured AI evidence returned by the SDK in the evaluation payload, with commit-pinned source URLs when repository metadata is available.
- Preserves the existing `NeedsReview` result and message when AI is not configured.
- Degrades provider, configuration, evidence-fetch, and response failures to `NeedsReview` instead of producing a false negative.
- Distinguishes an absent README from transient fetch or decode failures so partial evidence is not judged as complete.

## Supporting changes

- Adds `gemara.HasEvidence` support to `data.Payload` and initializes its evidence collector in the loader.
- Adds `RestData.GetFileContent` so callers can distinguish missing content from retrieval failures.
- Adds a failing REST client test helper for cross-package error-path tests.
- Adds a reusable AI fallback helper that logs the failure reason and returns manual review.

## Backward compatibility

When AI is not configured, the evaluator returns the same `NeedsReview` result and message as the previous implementation. AI and evidence failures also fall back to that result.

## Validation

- `make test`
- `gofmt`
- `git diff --check`
- GitHub CI, build, lint, CodeQL, DCO, and Kusari checks pass

PR #364, opened 2026-07-13
feat(OSPS-BR-01.03): detect untrusted code accessing privileged CI/CD credentials

### LINKED ISSUE #335: Add OSPS-BR-01.03 support
OSPS-BR-01.03 is present in OSPS Baseline 2026-02-19 but is not fully supported by the scanner yet.

Work needed:

- Add the requirement to the catalog data if it is missing
- Replace the current NotImplemented stub with a real check
- Add focused tests for the new check

Related context: older control-level issue ossf/pvtr-github-repo-scanner#16 was closed before this sub-requirement existed.

vinayada1:
## Summary
Implements **OSPS-BR-01.03** — *"When a CI/CD pipeline operates on untrusted code snapshots, it MUST prevent access to privileged CI/CD credentials and assets."* Closes #335.

- Adds `OSPS-BR-01.03` to the **2026-02** catalog (the baseline version that introduced this requirement), with the official recommendation text.
- Maps it to a new step, `build_release.CicdUntrustedCodeIsolation`.

## What it detects
A privileged workflow is **Failed** when both hold:
1. It runs in a **privileged context** — `pull_request_target`, `workflow_run`, or `issue_comment` (base-repo secrets + write token, but initiable by an untrusted actor).
2. It **checks out untrusted code** — `actions/checkout` with an untrusted head ref, or a run step doing `gh pr checkout` / `git checkout|fetch|switch` of a PR head.

False-positive guards: non-privileged `pull_request` + `gh pr checkout`, `git checkout main`, and base-SHA checkouts are **not** flagged.

## Result semantics (tiered for complete coverage)
To fully cover the requirement, a privileged workflow is never silently passed:

| Result | Meaning |
| --- | --- |
| `Failed` | Privileged workflow checks out untrusted code (the "pwn request" pattern). |
| `NeedsReview` | Workflow runs in a privileged context but no dangerous checkout was detected; the residual vectors below are not statically decidable, so a human must confirm isolation. |
| `Passed` | No workflow runs in a privileged context. |
| `NotApplicable` | Repository has no workflows. |

## 📝 Note to reviewers
Per #335 we aimed to **completely cover** BR-01.03. Deterministic detection alone would leave a silent-`Pass` gap: a privileged workflow whose exposure comes from a vector static analysis can't clear would be marked compliant. To close that gap without over-claiming, privileged-but-not-pattern-matched workflows now return **`NeedsReview`** (consistent with how the project already treats partially-automatable controls) rather than `Passed`.

Trade-off: repositories legitimately using `pull_request_target` / `workflow_run` / `issue_comment` will now surface for manual review instead of passing automatically. This is intentional and more thorough, but noisier — flagging it here for visibility. Happy to dial this back to a plain `Passed` if you'd prefer lower noise.

## Residual vectors handled by the `NeedsReview` tier
- **`workflow_run` artifact/cache poisoning** — a contextual dataflow judgment; candidate for an AI-assisted escalation layer on the `sdkai` seam (after #346).
- **Self-hosted runner on fork PRs** — depends on runner/fork settings not in the workflow file; needs an API-backed data source. (AI cannot infer settings it can't see, so this stays API-backed.)


data/catalogs/OSPS_Baseline_2025_10.yaml
```diff
@@ -549,6 +549,15 @@ controls:
...
         recommendation: # TODO
+      - id: OSPS-BR-01.03
+        text: |
+          When a CI/CD pipeline operates on untrusted code snapshots, it MUST
+          prevent access to privileged CI/CD credentials and assets.
+        applicability:
+          - Maturity Level 1
+          - Maturity Level 2
+          - Maturity Level 3
+        recommendation: # TODO
```
eddie-knight: This requirement was added in version 2026.02.19

```suggestion
```

data/catalogs/OSPS_Baseline_backwards-compatibility-protection.yaml
```diff
@@ -549,6 +549,15 @@ controls:
...
         recommendation: # TODO
+      - id: OSPS-BR-01.03
+        text: |
+          When a CI/CD pipeline operates on untrusted code snapshots, it MUST
+          prevent access to privileged CI/CD credentials and assets.
+        applicability:
+          - Maturity Level 1
+          - Maturity Level 2
+          - Maturity Level 3
+        recommendation: # TODO
```
eddie-knight:
```suggestion
```

data/catalogs/OSPS_Baseline_2026_02.yaml
```diff
@@ -549,6 +549,15 @@ controls:
...
         recommendation: # TODO
+      - id: OSPS-BR-01.03
+        text: |
+          When a CI/CD pipeline operates on untrusted code snapshots, it MUST
+          prevent access to privileged CI/CD credentials and assets.
+        applicability:
+          - Maturity Level 1
+          - Maturity Level 2
+          - Maturity Level 3
+        recommendation: # TODO
```
eddie-knight: The recommendation is available here: https://baseline.openssf.org/versions/2026-02-19.html#osps-br-0103

```suggestion
        recommendation: |
          CI/CD pipelines should isolate untrusted code snapshots from privileged credentials and assets. 
          In particular, projects should be careful to ensure that workflows which build or execute code 
          prior to review by a collaborator do not have access to CI/CD credentials.
```

eddie-knight (changes requested):
I still need to dig deeper into `checkWorkflowForUntrustedCodeAccess`, but in the meantime here are a couple quick feedback points

jmeridth (commented):
:robot: Multi-model review (Opus 4.8, Sonnet 5, Fable 5). Detection core and the 2026-02 catalog entry look solid — the notes below are all in the `CicdUntrustedCodeIsolation` file-walking loop, which diverges from the `evaluateWorkflows` contract. Three line comments follow; detection-completeness gaps (repository: input, env/matrix indirection, two-statement checkout, reusable-workflow calls) all safely degrade to NeedsReview so I've left them out as non-blocking.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -120,6 +147,194 @@ func CicdBranchNameSanitized(payload data.Payload) (gemara.Result, string, gemar
...
+	for _, file := range workflows {
+		if !strings.HasSuffix(file.Name, ".yml") && !strings.HasSuffix(file.Name, ".yaml") {
+			continue
+		}
+		if file.Truncated {
+			return gemara.NeedsReview, fmt.Sprintf("Workflow file %v is too large to retrieve for inspection; manual review required", file.Path), confidence
+		}
+		workflow, parseErr := actionlint.Parse([]byte(file.Content))
+		if parseErr != nil {
+			return gemara.Failed, fmt.Sprintf("Error parsing workflow: %v (%s)", parseErr, file.Path), confidence
```
jmeridth: :robot: (high) A parse failure here returns `Failed`, asserting the repo violates BR-01.03 based on a file we never understood — the exact thing sibling `evaluateWorkflows` deliberately avoids ("Failed asserts that the repository violates the control, which we have not observed for a file we never read"). Any workflow with an actionlint diagnostic and no privileged trigger gets reported as exposing credentials, and it short-circuits before scanning the rest. Suggest mirroring `evaluateWorkflows`: collect unparseable files as uninspected and fall through to `NeedsReview`.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -120,6 +147,194 @@ func CicdBranchNameSanitized(payload data.Payload) (gemara.Result, string, gemar
...
+		}
+		return gemara.NotApplicable, "No workflows found in .github/workflows directory", confidence
+	}
+
+	var parsed []namedWorkflow
+	for _, file := range workflows {
+		if !strings.HasSuffix(file.Name, ".yml") && !strings.HasSuffix(file.Name, ".yaml") {
+			continue
+		}
+		if file.Truncated {
```
jmeridth, line 187: :robot: (medium) This early-returns on the first truncated file, so a genuine pwn-request in a later workflow never gets classified — the verdict becomes order-dependent on GitHub's file ordering. `evaluateWorkflows` accumulates truncated files and keeps scanning so "unreadable siblings can never mask a real finding," and your own `TestClassifyUntrustedCodeIsolation` asserts `Failed` precedence at the classify layer. Recommend accumulating here and only degrading to `NeedsReview` after confirming no parsed workflow produced a violation.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -120,6 +147,194 @@ func CicdBranchNameSanitized(payload data.Payload) (gemara.Result, string, gemar
...
+
+// stepChecksOutUntrustedCode reports whether a run: script materializes an
+// untrusted PR head into the workspace. `gh pr checkout` always targets the PR
+// head, so it is flagged unconditionally; git checkout/fetch/switch is flagged
+// only when it also references an untrusted head ref, so a benign
+// `git checkout main` is not a false positive.
+func stepChecksOutUntrustedCode(script string) bool {
+	// Preserve line continuations before splitting the script into commands.
+	script = strings.ReplaceAll(script, "\\\n", " ")
+	for _, command := range shellCommandSeparator.Split(script, -1) {
```
jmeridth, line 356: :robot: (low, non-blocking) `stepChecksOutUntrustedCode` treats every split segment as a command, so comments/heredoc text defeat it in the false-positive direction. Verified against this exact logic: a privileged workflow whose run step contains `# git checkout pull/123/head is insecure` returns a violation -> authoritative `Failed`. Narrow, but a wrong `Failed` on a compliance control misreports the repo. Stripping `#`-comments before matching closes the common case.

evaluation_plans/osps/build_release/steps.go
vinayada1: Fixed: parse failures are now collected as uninspected and fall through to NeedsReview instead of Failed, mirroring evaluateWorkflows.

evaluation_plans/osps/build_release/steps.go
vinayada1: Fixed: truncated files are now accumulated and scanning continues, so a real Failed still takes precedence over an unreadable sibling regardless of ordering.

evaluation_plans/osps/build_release/steps.go
vinayada1: Fixed: shell #-comments are now stripped from each command segment before matching, so a checkout named only in a comment is no longer flagged.

eddie-knight (approved):

PR #365, opened 2026-07-14
feat(OSPS-BR-05.01): detect standardized dependency tooling

### LINKED ISSUE #338: Implement OSPS-BR-05.01
OSPS-BR-05.01 is currently wired as a stub. The scanner should replace it with a real check.

Work needed:

- Confirm the current OSPS Baseline 2026-02-19 requirement wording
- Implement the scanner check
- Add focused tests

vinayada1:
Implements OSPS-BR-05.01 using GitHub's dependency graph as a conservative signal for standardized dependency tooling.

- Replaces the assessment stub with a check that passes when GitHub detects dependency manifests and returns `NeedsReview` when none are detected.
- Includes detected manifest filenames as assessment evidence when available; filename lookup is optional and cannot abort a scan.
- Adds focused unit coverage for assessment results and GraphQL filename extraction/error handling.

Fixes #338


evaluation_plans/osps/build_release/steps.go
```diff
@@ -379,3 +379,21 @@ func SecretScanningInUse(payload data.Payload) (result gemara.Result, message st
...
+// available. GitHub's dependency graph only detects manifests produced by
+// standardized ecosystem tooling (e.g. go.mod, package.json, requirements.txt,
+// Cargo.toml, pom.xml), so the presence of at least one detected manifest is a
+// strong signal that dependencies are ingested through standardized tooling.
+func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	manifestsCount := payload.DependencyManifestsCount
+	if manifestsCount > 0 {
+		message := fmt.Sprintf("Found %d dependency manifest(s) in the GitHub dependency graph, indicating dependencies are ingested via standardized tooling", manifestsCount)
+		if len(payload.DependencyManifestFilenames) > 0 {
+			message += ": " + strings.Join(payload.DependencyManifestFilenames, ", ")
```
jmeridth: :robot: The count here is GitHub's unbounded `TotalCount`, but `DependencyManifestFilenames` is capped at `first: 100` (`graphql-dependency-manifests.go:54`). For a repo with >100 detected manifests, this renders e.g. "Found 150 dependency manifest(s)…: " followed by only 100 names — the evidence contradicts its own count. The truncation is noted in the `Payload` struct comment but never surfaced here. Consider either appending a "(showing first 100)" caveat when `len(filenames) < count`, or capping the stated count to what's listed.

data/payload.go
```diff
@@ -44,6 +49,14 @@ func Loader(config *config.Config) (payload any, err error) {
 		return nil, err
 	}
 
+	// Manifest filenames only enrich evidence messages, so a failure here
+	// degrades gracefully (empty list) rather than aborting the whole scan.
+	dependencyManifestFilenames, err := getDependencyManifestFilenames(client, config)
```
jmeridth: :robot: This is a second GraphQL round-trip (with its own retry) against the same `dependencyGraphManifests` field that `countDependencyManifests` just queried at line 47. A single query can return both — `dependencyGraphManifests(first: 100) { totalCount nodes { filename } }` — which removes the extra call and makes the count and filename list atomic (right now they're sampled at different instants and can diverge). Worth folding into one query.

data/graphql-dependency-manifests.go
```diff
@@ -42,3 +42,42 @@ func countDependencyManifests(client *githubv4.Client, cfg *config.Config) (int,
...
+// evidence. The authoritative count comes from countDependencyManifests.
+func getDependencyManifestFilenames(client *githubv4.Client, cfg *config.Config) ([]string, error) {
+	var query dependencyManifestFilenamesPage
+	variables := map[string]any{
+		"owner": githubv4.String(cfg.GetString("owner")),
+		"name":  githubv4.String(cfg.GetString("repo")),
+	}
+
+	err := withRetry(cfg.Logger, "GraphQL dependency manifest filenames query", func() error {
+		query = dependencyManifestFilenamesPage{}
```
eddie-knight: Is this line redundant...?

```suggestion
```

evaluation_plans/osps/build_release/steps.go
```diff
@@ -379,3 +379,21 @@ func SecretScanningInUse(payload data.Payload) (result gemara.Result, message st
...
+		}
```
eddie-knight: Any more than a handful of filenames here will make the runtime pretty long. Perhaps we could put it in a debug log? Or if we can't think of a specific value this info creates, we could just remove it.


```suggestion
```

eddie-knight (dismissed):
This looks good to me. I added two nitpick suggestions, but not blocking on them.

data/payload.go
vinayada1: Good catch — I folded this in while merging `main`. Since #373 ("cut GitHub API calls per scan from 20 to 9") the manifest count is already returned by the primary repo query (`DependencyGraphManifests { TotalCount }`), so I removed the separate filename query entirely and now report the count only. That collapses this to a single query at zero extra API cost and makes the count atomic (no divergence between two round-trips). It also resolves the >100 `TotalCount`-vs-capped-filenames concern on the other thread, since filenames are no longer listed.

evaluation_plans/osps/build_release/steps.go
vinayada1: Resolved by the count-only change: filenames are no longer appended to the message, so the unbounded `TotalCount` can no longer disagree with a capped list — the evidence now reports just the count from the primary repo query.

data/graphql-dependency-manifests.go
vinayada1: Removed — the separate filename query (and this whole file) is gone now that we report the manifest count only, so this line no longer exists.

evaluation_plans/osps/build_release/steps.go
vinayada1: Went with removing them: the evidence now reports just the manifest count, with no filename list.

jmeridth (approved):
:robot: Approving. Fable once-over on top of my and @eddie-knight's earlier reviews — all four threads resolved by the count-only refactor.

The manifest count now comes from the primary repo query (`DependencyGraphManifests.TotalCount`), no separate filename round-trip, and the partial-data/permission path zeroes to `NeedsReview` rather than a silent pass. Tiering (Passed on >0, NeedsReview on 0) matches the sibling QA step, docstring is accurate, and the table tests assert the messages verbatim. CI green.

PR #366, opened 2026-07-14
feat(OSPS-AC-04.02): check workflow permissions

### LINKED ISSUE #339: Implement OSPS-AC-04.02
OSPS-AC-04.02 is currently wired as a stub. The scanner should replace it with a real check.

Work needed:

- Confirm the current OSPS Baseline 2026-02-19 requirement wording
- Implement the scanner check
- Add focused tests

vinayada1:
## Summary

- replace the OSPS-AC-04.02 stub with a GitHub Actions workflow permissions check
- fail unambiguously excessive `write-all` grants
- request manual review for `read-all` and scoped grants whose necessity depends on job behavior
- pass explicit no-access configurations and report no explicit assignment as not applicable
- add focused tests for workflow-level and job-level permission configurations

Fixes #339


evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+	var reviewRequired []string
+	permissionsAssigned := false
+
+	for _, file := range workflows {
+		if !strings.HasSuffix(*file.Name, ".yml") && !strings.HasSuffix(*file.Name, ".yaml") {
+			continue
+		}
+
+		if *file.Encoding != "base64" {
+			return gemara.Failed, fmt.Sprintf("File %v is not base64 encoded", *file.Name), confidence
```
eddie-knight:
```suggestion
			continue
```

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+		}
+
+		decoded, decodeErr := base64.StdEncoding.DecodeString(*file.Content)
+		if decodeErr != nil {
+			return gemara.Failed, fmt.Sprintf("Error decoding workflow file: %v", decodeErr), confidence
```
eddie-knight:
```suggestion
			continue
```

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+		}
+
+		workflow, parseErr := actionlint.Parse(decoded)
+		if parseErr != nil {
+			return gemara.Failed, fmt.Sprintf("Error parsing workflow: %v (%s)", parseErr, *file.Path), confidence
```
eddie-knight:
```suggestion
			continue
```

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+		if *file.Encoding != "base64" {
+			return gemara.Failed, fmt.Sprintf("File %v is not base64 encoded", *file.Name), confidence
+		}
+
+		decoded, decodeErr := base64.StdEncoding.DecodeString(*file.Content)
+		if decodeErr != nil {
+			return gemara.Failed, fmt.Sprintf("Error decoding workflow file: %v", decodeErr), confidence
+		}
+
+		workflow, parseErr := actionlint.Parse(decoded)
```
jmeridth: :robot: This hand-rolls what go-github already does. `(*RepositoryContent).GetContent()` handles the `base64` / `none` / empty encodings for you — with a nil-content guard and a descriptive error on the >1 MB `none` case — and it's already the idiom used in `quality/steps.go`. Using it here also removes the two `*file.Encoding` / `*file.Content` nil dereferences.

```suggestion
		decoded, err := file.GetContent()
		if err != nil {
			return gemara.Failed, fmt.Sprintf("Error reading workflow file %s: %v", *file.Path, err), confidence
		}

		workflow, parseErr := actionlint.Parse([]byte(decoded))
```

One knock-on a suggestion can't reach: this removes the only use of `encoding/base64`, so drop that import at the top of the file.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+		}
+		assigned = true
+		if perms.All != nil {
+			if strings.EqualFold(perms.All.Value, "write-all") {
+				violations = append(violations, label+" grant write-all")
+			} else {
+				reviewRequired = append(reviewRequired, fmt.Sprintf("%s grant %s", label, perms.All.Value))
+			}
+			return
+		}
```
jmeridth, line 371: :robot: Only the `write-all` shorthand reaches `Failed` here. A job that grants every scope individually as `write` (`contents: write`, `packages: write`, `id-token: write`, ...) is effectively write-all but falls through to the `Scopes` loop below and only gets `NeedsReview`. That lets the most permissive long-form config dodge the unambiguous fail this control is meant to raise. If that's a deliberate choice (scoped grants -> manual review), worth a comment saying so; otherwise consider failing when a block grants `write` across all/most scopes.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+		for scope, permission := range perms.Scopes {
+			if permission.Value != nil && !strings.EqualFold(permission.Value.Value, "none") {
+				reviewRequired = append(reviewRequired,
+					fmt.Sprintf("%s grant %s: %s", label, scope, permission.Value.Value))
+			}
+		}
+	}
+
+	// Top-level permissions apply to every job that does not override them.
+	check(workflow.Permissions, fmt.Sprintf("%s: workflow-level permissions", name))
```
jmeridth: :robot: Workflow-level and job-level `permissions:` are checked independently, but GitHub *replaces* the workflow-level block when a job defines its own (it doesn't merge). So a permissive top-level block that every job overrides with something stricter still gets flagged here — e.g. `permissions: write-all` up top with the only job overriding to `{contents: read}` would still report `Failed` on the now-dead top-level grant. It only ever over-reports (never misses a real problem), but modeling the override — skip the workflow-level check for jobs that define their own block — would cut false positives.

evaluation_plans/osps/access_control/steps_test.go
```diff
@@ -284,3 +285,127 @@ func Test_BranchProtectionPreventsDeletion(t *testing.T) {
 		})
 	}
 }
+
+func Test_checkWorkflowJobPermissions(t *testing.T) {
```
jmeridth, line 398: :robot: Nice coverage of `checkWorkflowJobPermissions`. The outer `WorkflowJobPermissionsLeastPrivilege` is untested though — the aggregation, the abort-on-parse/decode-error paths, and the `NotApplicable` / `Passed` / multi-file branches have no tests. A couple of table cases driving the exported function with a mocked `data.Payload` (a good workflow, a `write-all` one, and a malformed file) would lock down the failure modes.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -70,3 +76,145 @@ func WorkflowDefaultReadPermissions(payload data.Payload) (result gemara.Result,
...
+
+	if len(reviewRequired) > 0 {
+		return gemara.NeedsReview,
+			"CI/CD job permissions require review to confirm they are necessary: " + strings.Join(reviewRequired, "; "),
+			confidence
+	}
+
+	if !permissionsAssigned {
+		return gemara.NotApplicable, "No CI/CD jobs explicitly assign permissions", confidence
+	}
```
eddie-knight, line 307: Bumping the NotApplicable up higher would improve readability (and save a few microseconds)

```suggestion
	if !permissionsAssigned {
		return gemara.NotApplicable, "No CI/CD jobs explicitly assign permissions", confidence
	}

	sort.Strings(violations)
	sort.Strings(reviewRequired)

	if len(violations) > 0 {
		return gemara.Failed,
			"CI/CD jobs assign more than the minimum privileges: " + strings.Join(violations, "; "),
			confidence
	}

	if len(reviewRequired) > 0 {
		return gemara.NeedsReview,
			"CI/CD job permissions require review to confirm they are necessary: " + strings.Join(reviewRequired, "; "),
			confidence
	}
```

evaluation_plans/osps/access_control/steps.go
vinayada1: Agreed on the long-form case — `permissionsGrantMaximumAccess` now fails a block that grants every scope at max, same as `write-all`, at both workflow and job level.

I left the "most scopes" idea out on purpose: there is no non-arbitrary threshold, and legit jobs (e.g. a release job with `contents` + `packages` + `id-token`) often need several `write` scopes. Those stay `NeedsReview` for a human to judge against the job's activity.

evaluation_plans/osps/access_control/steps.go
vinayada1: Agreed on the readability goal, but I kept `NotApplicable` below the violations/uninspected checks on purpose. When a workflow cannot be read (malformed or too large), the `permissionsAssigned` flag stays false — same as a file with no permissions. Moving the `NotApplicable` check to the top would make those unreadable files return `NotApplicable` and swallow the `NeedsReview` signal a human needs. So the order is: no workflows -> `NotApplicable`, violations -> `Failed`, unreadable -> `NeedsReview`, then no-permissions -> `NotApplicable`.

jmeridth (approved):
:robot: Approving. Fable once-over on top of my and @eddie-knight's earlier reviews — all eight threads resolved and verified fixed in code, not just marked.

Confirmed the four substantive points landed: long-form write-all now fails via `permissionsGrantMaximumAccess` (scope map matches pinned actionlint, `models: read` correct); the dead workflow-level grant is skipped when every job overrides it (`workflowPermissionsApply`); the exported/eval functions now have table coverage including parse and truncation paths; and the result ordering (violations before uninspected before NotApplicable) means a malformed sibling can neither mask a real Failed nor be swallowed into NotApplicable. Tests green, CI green.

PR #368, opened 2026-07-16
chore: upgrade sdk to v1.31.3, improves Gemara output logs

eddie-knight:
What it says on the tin.


jmeridth (approved):

jmeridth:
Nice. Valid catches it seems.

jmeridth:
Dev binaries are removed in https://github.com/ossf/pvtr-github-repo-scanner/pull/346 (as you pointed out earlier today in the meeting where we looked at that PR)

PR #373, opened 2026-07-20
chore: cut GitHub API calls per scan from 20 to 9

eddie-knight:
| Previous Code Benchmark | Updated Benchmark |
| --- | --- |
| 6.83s | 6.34s |
| 7.18s | 5.21s |
| 7.71s | 5.45s |
| 7.85s | 5.11s |
| 6.49s | 4.79s |

Three sources of avoidable requests, in descending cost:

Workflow files were fetched one REST call per file after a directory listing, the only unbounded API cost in a scan. A single GraphQL query now selects each entry's name, path, and text inline.

Dependency manifest counts had their own GraphQL query even though they target the same repository node as the main one, so they are selected alongside it at no extra cost.

Branch rulesets were fetched over REST despite already being loaded into RepositoryMetadata during payload load. HasBranchRules and RequiredStatusCheckContexts read the loaded copy instead.

Two caching fixes support this. getSubdirContents now writes fetched directories back, so repeat probes of the same directory (checkFile alone looks in .github once per filename) hit the cache. Payload caches live behind a pointer because steps receive the Payload by value, so a plain field would be written to a per-step copy and refetched.

Removes GetDirectoryContent, GetFileContent, GetSubdirContentByPath, GetRulesets, and the Ruleset type, which have no remaining callers.

Note for --include-payload consumers: the RestData.Rulesets field is gone. It was only ever populated by GetRulesets, so retaining it would emit a permanently empty array.


data/graphql-workflows.go
```diff
@@ -0,0 +1,96 @@
...
+	return workflowFilesFromQuery(query, cfg.Logger), nil
+}
+
+// workflowFilesFromQuery maps a directory listing onto WorkflowFiles, dropping
+// entries whose contents cannot be evaluated.
+func workflowFilesFromQuery(query GraphqlWorkflowFiles, logger hclog.Logger) []WorkflowFile {
+	var files []WorkflowFile
+	for _, entry := range query.Repository.Object.Tree.Entries {
+		// Subdirectories, submodules, and symlinks carry no blob to inspect.
+		if entry.Type != "blob" || entry.Object == nil {
```
jmeridth: :robot: (low-medium) In git's object model a symlink is a blob (mode `120000`) whose text is the link-target path, and GraphQL reports `type: "blob"` for it, so this skip doesn't catch symlinks and the comment's "symlinks carry no blob to inspect" isn't quite right. A symlinked `.github/workflows/ci.yml` comes back with `Content` set to the target path (or `""` for a non-UTF-8 blob, since `Blob.text` is null). `checkAllWorkflows` then hits `actionlint.Parse`, which errors (`"workflow" section is scalar node...` / `workflow is empty`) and hard-`Failed`s the whole control. The old REST path skipped these via `GetType() != "file"` (symlinks report as `"symlink"`), so this is a false-positive regression for repos that symlink workflows. Verified by running `actionlint.Parse` on both a symlink target path and empty bytes: both return errors. Selecting the entry `mode` and skipping `120000` (or skipping when `Content` is empty) would restore the old skip behavior.

_AI-generated review (Opus 4.8, cross-checked with Sonnet 5 / Haiku 4.5 / Fable 5)._

data/graphql-workflows.go
```diff
@@ -0,0 +1,96 @@
...
+			continue
+		}
+		// GitHub truncates very large blobs. Parsing a partial workflow would
+		// produce misleading results, so skip it and say so.
+		if entry.Object.Blob.IsTruncated {
+			logger.Warn(fmt.Sprintf("skipping truncated file, too large to evaluate: %s", entry.Path))
```
jmeridth: :robot: (low, non-blocking) A GraphQL-truncated (oversized) workflow is dropped here with only a Warn, so `checkAllWorkflows` evaluates the remaining files and can still return `Passed` without ever inspecting this one. The old REST path errored on oversized content and returned `NotApplicable`, so this is the one fail-open direction in the new path. It needs a >~512KB workflow YAML, which is implausible, and it's unclear Actions would even run such a file, so this is just a flag rather than a blocker. Degrading to `NeedsReview` when any workflow is truncated would close it.

_AI-generated review (Opus 4.8, cross-checked with Sonnet 5 / Haiku 4.5 / Fable 5)._

data/graphql-workflows.go
eddie-knight: Good catch — fixed in b59e5eb by skipping mode 120000. The parser never sees symlinks now.

data/graphql-workflows.go
eddie-knight: Agreed. Deferring this to a follow-up PR to keep this one focused on the symlink regression — not addressed here.

jmeridth (approved):

PR #376, opened 2026-07-21
perf: parallelize payload retrieval stages

eddie-knight:
What it says on the tin.

Using `waitgroup` instead of `errorgroup` because every one of these is safely nullable and resolved later in the assessments.

Benchmarking the payload retrieval specifically:

| `main` | `perf/parallelize-loader` |
| --- | --- |
| 3.68s | 1.48s |
| 3.22s | 1.62s |
| 2.93s | 1.33s |
| 2.84s | 1.93s |
| 4.00s | 1.64s |

## Unintended Consequences

This also fixes a bug related to branch protection data on user branches, which would always fail previously.

### DIFF
--- a/data/payload.go
+++ b/data/payload.go
@@ -10,6 +10,7 @@ import (
 	"github.com/privateerproj/privateer-sdk/pluginkit"
 	"github.com/shurcooL/githubv4"
 	"golang.org/x/oauth2"
+	"golang.org/x/sync/errgroup"
 )
 
 type Payload struct {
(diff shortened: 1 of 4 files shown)


data/rest-data.go
```diff
@@ -63,14 +64,27 @@ type WorkflowPermissions struct {
...
 func (r *RestData) Setup() error {
-	r.owner = r.Config.GetString("owner")
-	r.repo = r.Config.GetString("repo")
-	r.token = r.Config.GetString("token")
+	// owner/repo/token are resolved by newRestData
+	if r.owner == "" && r.Config != nil {
+		r.owner = r.Config.GetString("owner")
+		r.repo = r.Config.GetString("repo")
+		r.token = r.Config.GetString("token")
+	}
```
jmeridth: :robot: (Low-Med, non-blocker) This guard is a latent data race. `Setup` and `IsCodeRepo` run concurrently on the same `*RestData` (`payload.go:78-84`). When the `owner` config value is empty, this branch writes `r.owner/r.repo/r.token` while `IsCodeRepo` concurrently reads `r.owner`/`r.repo` — an unsynchronized torn read. `-race` misses it only because no test drives an empty owner. Since `newRestData` already resolves these fields and is the only path that reaches `Setup`, this fallback is dead code in practice — dropping it makes the race structurally impossible.

```suggestion
	// owner/repo/token are resolved by newRestData; Setup is only ever
	// reached through it, so no fallback resolution is needed here.
```

data/payload.go
```diff
@@ -52,29 +53,40 @@ func Loader(config *config.Config) (payload any, err error) {
...
+	var (
+		graphql            *GraphqlRepoData
+		client             *githubv4.Client
+		ghRepo             *github.Repository
+		repositoryMetadata RepositoryMetadata
+		isCodeRepo         bool
+	)
+	rest := newRestData(ghClient, httpClient, config)
+
+	g := new(errgroup.Group)
```
jmeridth, line 69: :robot: (Scale note, non-blocker) Solid per-scan latency win. One thing worth capturing for the orchestration layer: this raises peak in-flight GitHub requests per scan from 1 to ~7 (this errgroup fans out 4 ways, and the metadata and Setup branches each fan out again). At bulk scale (LF Insights runs this against ~27.5k repos) the primary hourly budget — unchanged by this PR — is the real ceiling, so per-token pacing lands at ~1-2 concurrent scans anyway; the ~7x in-flight multiplier only matters once the run is spread across several App installations, where keeping concurrent scans under the 100-request secondary limit becomes the guard. Not a change request here — just so nobody sizes parallelism assuming 1 request per repo.

data/rest-data.go
```diff
@@ -63,14 +64,27 @@ type WorkflowPermissions struct {
...
-	r.loadSecurityInsights()
-	_ = r.getWorkflowPermissions()
-	_ = r.getReleases()
+
+	// Errors are expected when one of these are not in place; safe to ignore
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		r.loadSecurityInsights()
+	})
+	wg.Go(func() {
```
jmeridth, line 77: :robot: (Low, non-blocker) Now that `getWorkflowPermissions` and `getReleases` run concurrently, both reach `MakeApiCall`'s unsynchronized lazy init `if r.HttpClient == nil { r.HttpClient = &http.Client{} }` (`rest-data.go:99-101`). Not reachable via `Loader` (`newRestData` always sets `HttpClient`), so this is latent — but any future caller constructing `RestData` without a client would hit a check-then-write race. Cheap to pre-empt by setting the client in the constructor.

jmeridth (dismissed):
Some non-blocking comments. approving.

jmeridth (approved):

PR #385, opened 2026-07-21
fix: evaluate observable links for BR-03 when Security Insights is absent

eddie-knight:
Fixes OSPS-BR-03.

The link check was gated behind `HasSecurityInsightsFile`, so **1,701 of 1,882** repos went to `NeedsReview` for links that are directly observable.

### What changes
- `EnsureInsightsLinksUseHTTPS` handles SI-or-fallback itself: with SI, enumerate declared links; without it, check the observable set (repo homepage + release asset URLs) — Pass when all HTTPS (or none exist, since GitHub is HTTPS-only), Fail on any insecure link.
- `DistributionPointsUseHTTPS` uses release assets as observable distribution points; stays `NotApplicable` only when nothing is observable.
- Drop the `HasSecurityInsightsFile` guard from BR-03.01. Add a `Homepage` accessor.
- Test hygiene: the metadata mock now **embeds** the interface and overrides only what the links steps read — so it no longer breaks when another branch grows the interface (makes this PR order-independent of AC-03).

> **Rebase note:** shares `data/repository_metadata.go` (AC-03), `evaluation-plans.go`, `build_release/steps.go` (BR-04) — second to merge needs a trivial rebase.

### DIFF
--- a/data/repository_metadata.go
+++ b/data/repository_metadata.go
@@ -11,6 +11,7 @@ import (
 type RepositoryMetadata interface {
 	IsActive() bool
 	IsPublic() bool
+	Homepage() string
 	OrganizationBlogURL() *string
 	IsDefaultBranchProtected() *bool
 	DefaultBranchRequiresPRReviews() *bool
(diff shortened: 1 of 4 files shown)


jmeridth:
@eddie-knight conflict

jmeridth (approved):

PR #387, opened 2026-07-21
fix: scan the full repository tree for binaries with explicit truncation handling

eddie-knight:
**Fixes a real missed binary + unknown false-passes** on OSPS-QA-05, resolves 129 `Unknown`s, and cuts scan time **~4–18s → ~1–2s**.

The scan used a GraphQL tree hardcoded to **3 levels deep**: it errored out (→ `Unknown`) on huge repos and could **falsely pass** a repo whose only binaries live at depth 4+. A full-depth scan found a real committed binary the old one missed.

### What changes
- Replace the depth-3 GraphQL tree with a single `REST git/trees?recursive=1` — full depth plus an explicit `truncated` flag.
- REST carries no `IsBinary`/content, so binary status is inferred by extension (existing text/acceptable classifiers + a new binary-artifact list) feeding the unchanged `binaryChecker` — **still one call, no per-file fetches**.
- Honest coverage: a finding → **Fail** (even if truncated); clean over a **truncated** tree → **NeedsReview**; clean over a full tree → **Pass**; failed fetch → **Unknown**.

Net API calls unchanged (one GraphQL → one REST), shared across QA-05.01/.02 via the payload cache.

> **Rebase note:** shares `quality/steps.go` (QA-02) — second to merge needs a trivial rebase.

### DIFF
--- a/data/binary_scan_test.go
+++ b/data/binary_scan_test.go
@@ -1,0 +1,146 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests drive the QA-05 accessors through a pre-populated cache so no API
+// call is made. They cover the two behaviors the REST recursive tree adds:
+// full-depth detection and propagation of the truncation flag.
+
+func TestClassifyIsBinary(t *testing.T) {
+	tests := []struct {
+		path string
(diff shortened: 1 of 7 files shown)


eddie-knight:
Closing this as it causes more problems than it solves by moving away from the GitHub builtin inspection provided by `IsBinary`

PR #399, opened 2026-07-22
fix: reduce QA-05.02/05.01 binary-check false positives

eddie-knight:
Expand the acceptable-binary allowlist so genuinely reviewable, non-code assets are no longer flagged as unreviewable binary artifacts (OSPS-QA-05.02):

- Design/graphics sources: .ai, .sketch, .psd, .graffle
- Office documents: .docx, .doc, .pptx, .xlsx (same reviewability class as the already-excluded .pdf)

Also fix a QA-05.01 false positive: check() now skips acceptable binary extensions even when a Unix execute bit is set, so an image with a stray execute bit (a common artifact of checkouts from filesystems without Unix permissions) is not reported as a generated executable artifact.

Archives (.tar, .zip), JARs, and compressed scientific data remain flagged by design; those are genuinely unreviewable in-repo.


jmeridth:
@eddie-knight DCO failing

eddie-knight:
fixed

jmeridth (approved):

PR #401, opened 2026-07-22
fix: observability + result-semantics for QA-03/QA-04

eddie-knight:
- **fix: make QA-03 status-check evaluation observability-aware**
- **fix: report QA-04 as NeedsReview when Security Insights is absent**
- **docs: document assessment result semantics and evidence precedence**


jmeridth:
Nudge. Conflicts to fix.

PR #402, opened 2026-07-22
feat: AI-assist foundation + DO-01 user-guide assessment

eddie-knight:
- **feat: wire SDK AI client and evidence collector into the payload**
- **feat: assess DO-01 user guides from GitHub signals with AI assist**

### DIFF
--- a/data/payload.go
+++ b/data/payload.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/gemaraproj/go-gemara"
 	"github.com/google/go-github/v74/github"
+	"github.com/privateerproj/privateer-sdk/ai"
 	"github.com/privateerproj/privateer-sdk/config"
 	"github.com/privateerproj/privateer-sdk/pluginkit"
 	"github.com/shurcooL/githubv4"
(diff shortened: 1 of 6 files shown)


jmeridth:
Nudge. conflicts to fix.

eddie-knight:
🤖 Status note so this is recorded rather than rediscovered later. Branch is ~3 weeks stale and currently `CONFLICTING`; the two commits have aged in opposite directions.

**Commit 1 — "wire SDK AI client and evidence collector into the payload" is now a competing design, not just stale.** The AI foundation landed on main by a different route: `reusable_steps.AIFallback` plus a per-package `newAIClientFromConfig`, with no AI client on `data.Payload`. Three steps already use that pattern (`TestExecutionDocumentation`, `DocumentsTestMaintenancePolicy`, `WorkflowJobPermissionsLeastPrivilege`). Rebasing this commit would reintroduce a second way to do the same thing.

**Commit 2 — the DO-01 assessment is still novel.** `docs.HasUserGuides` on main is unchanged: six lines reading `Insights.Project.Documentation.DetailedGuide` and nothing else.

Suggested path when this is picked back up: drop commit 1, and reimplement the DO-01 assessment against the shipped pattern rather than resolving conflicts across all six files. That trades 400 lines over 6 files for one step function following three existing precedents.

Not acting on this now — flagging so the rebase isn't attempted on the assumption that only text conflicts are in the way.

PR #403, opened 2026-07-23
feat: implement OSPS-BR-07.02 secrets management policy check

vinayada1:
## Summary

Implements **OSPS-BR-07.02** — "The project MUST define a policy for managing secrets and credentials used by the project."

Previously this control was wired to `reusable_steps.NotImplemented`. This PR adds a real evaluation step.

## Changes

- **`data/security-posture.go`**: Add `definesSecretsHandlingPolicy()` which inspects the already-fetched `SECURITY.md` content. It requires both a mention of secrets/credentials **and** a management indicator (rotation, vault, secret/credential/key management) so that an incidental reference (e.g. "we scan for secrets") does not count. Wired into both `RepoSecurityPosture` construction paths via the new `definesPolicyForHandlingSecrets` field; removed the stale TODO.
- **`evaluation_plans/osps/build_release/steps.go`**: New `SecretsManagementPolicy` step. Returns `Passed` (Medium) when an observable policy is found, otherwise `NeedsReview` (Low) — the policy may live in documentation we cannot read (external wiki, handbook, etc.), so it is unconfirmed rather than a violation.
- **`evaluation_plans/evaluation-plans.go`**: Map `OSPS-BR-07.02` to `build_release.SecretsManagementPolicy`.
- Tests for the heuristic, the step (Passed/NeedsReview), and end-to-end `buildSecurityPosture`.

## Notes

- Detection is heuristic and intentionally conservative; when nothing is observable the result is `NeedsReview` to prompt manual confirmation rather than failing the control.

## Testing

- `go build ./...`, `go vet ./...`, and `go test ./data/... ./evaluation_plans/...` all pass.

### DIFF
--- a/data/security-posture.go
+++ b/data/security-posture.go
@@ -1,6 +1,8 @@
 package data
 
 import (
+	"strings"
+
 	"github.com/google/go-github/v74/github"
 	"github.com/ossf/si-tooling/v2/si"
 )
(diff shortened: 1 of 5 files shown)


eddie-knight (dismissed):

jmeridth:
@vinayada1 conflict to resolve here.

vinayada1:
Resolved the conflicts

data/security-posture.go
```diff
@@ -41,27 +43,62 @@ type RepoSecurityPosture struct {
...
+		secretScanningObservable:        true,
+		insightsDeclaresSecretScanning:  insightsClaimsSecretsTooling,
 	}, nil
 }
 
+// secretsPolicyIndicators are lowercase phrases that, alongside a mention of
+// secrets or credentials, signal a written policy for managing them (how they
+// are stored, accessed, or rotated) rather than an incidental reference such as
+// "we scan for secrets".
+var secretsPolicyIndicators = []string{
```
jmeridth, line 72: :robot: (low, non-blocking) These indicators plus the document-wide AND produce false Passes on this MUST control. Verified against the predicate: `rotat` collides with `log rotation`/on-call/release rotation and bare `vault` matches `vaulted`/metaphors, and since the `secret`/`credential` term and the indicator only need to co-occur anywhere in the file, a SECURITY.md with boilerplate `"Do not commit secrets"` plus a `"log rotation"` line returns `Passed` — as does `"we do not rotate credentials"` (no negation handling). The multi-word indicators (`secret manager`, `key management`) are safe; `rotat` and bare `vault` are the weak ones. Only error mode is over-passing (fallback is `NeedsReview`), so non-blocking, but scoping the AND to the same line/sentence — or using `key rotation`/`credential rotation` instead of bare `rotat` — would kill the common false Pass. A negation test case would document the limit.

jmeridth (approved):
:robot: Approving — multi-model review (Opus 4.8, Sonnet 5, Fable 5). First review on this PR.

Structurally clean and verified: wiring matches sibling BR-07.01 (ungated single step) and is contract-test-protected; the `DefinesPolicyForHandlingSecrets()` method/field already existed on main as a TODO stub and this PR correctly implements the setter in BOTH construction paths; the catalog entry exists in all three catalogs (ML3, text matches, no drift); load ordering populates SECURITY.md content before `buildSecurityPosture` reads it; no nil-panic; and never-returning-Failed is the right call here (no observable signal that a policy is absent — it may live in an external handbook), matching the repo's documented NeedsReview-over-Failed convention. Tests pass.

One low, non-blocking note left inline: the substring indicators plus a document-wide AND can false-Pass this MUST control (`log rotation` + boilerplate `do not commit secrets`, or a negated `we do not rotate credentials`). Since the only error mode is over-passing (fallback is NeedsReview), it's a bounded accuracy tradeoff, not a bug — but `rotat`/bare `vault` are worth tightening. Not blocking. Leaving the merge to you.

Issue #20, opened 2025-03-23
Implement OSPS-DO-05


eddie-knight:
The following test requirements are not yet implemented:

- [ ] When the project has made a release, the project documentation MUST provide a descriptive statement when releases or versions will no longer receive security updates.

eddie-knight:
Brilliant, thanks @trumant! 

I'm inclined to think this is another security insights check, but happy to leave it open to your creativity if you have another way of evaluating this one

PR #405, opened 2026-07-27
feat: implement OSPS-DO-05.01 security-update support statement check

### LINKED ISSUE #341: Implement OSPS-DO-05.01
OSPS-DO-05.01 is still a NotImplemented stub even though the older control-level issue was closed.

Work needed:

- Replace the stub with a real check
- Add focused tests
- Link back to ossf/pvtr-github-repo-scanner#20 for history

vinayada1:
## Why

`OSPS-DO-05.01` was still a `NotImplemented` stub even though the control-level work was tracked and closed long ago. The control requires that, once a project has made a release, its documentation states when releases or versions will stop receiving security updates.

## What changed

Replaced the stub with a real assessment step, `DocumentsSecurityUpdatePolicy`, in the existing `docs` package. It uses the Security Insights `Project.Documentation.SupportPolicy` field as the primary, machine-readable signal (that field maps exactly to "how releases are supported / end-of-life timelines" and was previously unused), and wires it into `evaluation-plans.go`.

Result logic, following conventions already used by the other docs and release steps:

- No releases -> `NotApplicable` (the requirement is explicitly conditional on a release existing, mirroring `EnsureLatestReleaseHasChangelog`).
- `SupportPolicy` set in Security Insights -> `Passed` (High confidence).
- No SI field but a `SUPPORT.md` / README "Support" section exists -> `NeedsReview` (Medium), since the file's contents cannot be verified to describe a security-update timeline.
- Nothing found -> `Failed` (Medium).

## Notes for reviewers

The `NeedsReview` fallback is intentional: a support document may exist without actually stating an end-of-life policy, so we flag it for a human rather than passing or failing outright. This matches the confidence philosophy in `AcceptsVulnReports`.

## Tests

Added a table-driven `TestDocumentsSecurityUpdatePolicy` covering all four outcomes, using the existing repo-contents mock. `go build ./...`, `go vet ./evaluation_plans/...`, and `go test ./evaluation_plans/...` all pass.

Relates to ossf/pvtr-github-repo-scanner#20 for history.

Fixes: #341

### DIFF
--- a/evaluation_plans/osps/docs/steps.go
+++ b/evaluation_plans/osps/docs/steps.go
@@ -68,6 +68,31 @@ func HasDependencyManagementPolicy(payload data.Payload) (result gemara.Result, 
 	return gemara.Failed, "Dependency management policy was NOT specified in Security Insights data", gemara.Medium
 }
 
+// DocumentsSecurityUpdatePolicy evaluates OSPS-DO-05.01: once a project has made
+// a release, its documentation must describe when releases or versions will no
+// longer receive security updates.
+//
+// The requirement is conditional on a release existing, so a project with no
+// releases is NotApplicable rather than a failure. Security Insights'
+// support-policy field is the explicit, machine-readable signal; a SUPPORT.md
+// (or README "Support" section) is a weaker fallback that warrants review since
+// its contents cannot be verified to describe a security-update timeline.
+func DocumentsSecurityUpdatePolicy(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	if len(payload.Releases) == 0 {
+		return gemara.NotApplicable, "No releases found; the security-update support statement requirement does not apply", confidence
+	}
+
+	if payload.Insights.Project.Documentation.SupportPolicy != nil {
+		return gemara.Passed, "Security update support policy was specified in Security Insights data", gemara.High
+	}
+
+	if payload.HasSupportMarkdown() {
+		return gemara.NeedsReview, "No support-policy field in Security Insights, but a SUPPORT.md file or a Support section in the readme.md was found; manual review required to confirm it states when releases stop receiving security updates", gemara.Medium
+	}
+
+	return gemara.Failed, "No security update support policy was found in Security Insights data and no SUPPORT.md file or Support section in the readme.md was found", gemara.Medium
+}
+
 func HasIdentityVerificationGuide(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
 	if payload.Insights.Project.Documentation.SignatureVerification == nil {
 		return gemara.Failed, "Identity verification guide was NOT specified in Security Insights data (checked signature-verification field)", confidence
--- a/evaluation_plans/osps/docs/steps.go
+++ b/evaluation_plans/osps/docs/steps.go
@@ -1,6 +1,8 @@
 package docs
 
 import (
+	"strings"
+
 	"github.com/gemaraproj/go-gemara"
 	"github.com/ossf/pvtr-github-repo-scanner/data"
 )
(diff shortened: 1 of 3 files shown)


evaluation_plans/osps/docs/steps.go
```diff
@@ -68,6 +68,31 @@ func HasDependencyManagementPolicy(payload data.Payload) (result gemara.Result,
...
+// releases is NotApplicable rather than a failure. Security Insights'
+// support-policy field is the explicit, machine-readable signal; a SUPPORT.md
+// (or README "Support" section) is a weaker fallback that warrants review since
+// its contents cannot be verified to describe a security-update timeline.
+func DocumentsSecurityUpdatePolicy(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	if len(payload.Releases) == 0 {
+		return gemara.NotApplicable, "No releases found; the security-update support statement requirement does not apply", confidence
+	}
+
+	if payload.Insights.Project.Documentation.SupportPolicy != nil {
```
jmeridth: :robot: (low, non-blocking) Verified against `si.Load` v2.2.0: a present-but-empty `support-policy: ""` decodes to a non-nil `*URL` of `""`, so this `!= nil` check reports `Passed` (High) for an empty field. Not introduced here — every sibling (`HasIdentityVerificationGuide`, `HasSignatureVerificationGuide`, `HasDependencyManagementPolicy`) uses the same `!= nil` idiom, so it's a package-wide convention. If you want to tighten it here:
```suggestion
	if sp := payload.Insights.Project.Documentation.SupportPolicy; sp != nil && strings.TrimSpace(string(*sp)) != "" {
```
(needs a `strings` import if not already present).

jmeridth (commented):
:robot: Multi-model review (Opus 4.8, Sonnet 5, Fable 5). Clean and consistent: SupportPolicy is the correct primary signal (SI schema maps it to EOL/support timelines), nil-safety is guarded by ensureInsightsInitialized, tiering matches siblings, wiring is contract-test-protected, catalog has no drift, and no Copilot hallucinations. Tests pass and were mutation-checked. One low/non-blocking note below.

jmeridth (dismissed):
:robot: Approving — multi-model review (Opus 4.8, Sonnet 5, Fable 5).

Clean and consistent across the board. The SI `support-policy` field is the correct primary signal (its schema doc maps to EOL/security-update timelines, citing endoflife.date), the double-pointer deref is safe because `ensureInsightsInitialized` guarantees non-nil `Project`/`Documentation` before any step runs, and the four-way tiering matches siblings: `NotApplicable`-on-no-releases mirrors `EnsureLatestReleaseHasChangelog`, `Passed/High` on the SI field mirrors `AcceptsVulnReports`, and `Failed/Medium` is right since all three signals are directly observable. The `SUPPORT.md → NeedsReview` fallback is justifiably more conservative than DO-06 since file presence does not prove a security-update timeline is stated. Wiring is contract-test-protected, the catalog has no drift, the docstring is accurate, and nothing was hallucinated by the Copilot co-authoring. Tests pass and were mutation-checked.

One low, non-blocking note left inline: a present-but-empty `support-policy: ""` decodes to a non-nil empty `*URL` and reports Passed — but that is the same `!= nil` idiom every sibling uses, so it is a package-wide convention rather than a regression. Fine to tighten here or leave.

Not blocking on anything. Approving; leaving the merge to you.

evaluation_plans/osps/docs/steps.go
vinayada1: Good catch. Tightened the check to trim and reject an empty `support-policy: ""` so an empty field now falls through to the SUPPORT.md fallback / Failed path instead of reporting Passed, and added a test for it (c83555a). Left the sibling steps as-is since this only affects DO-05.01.

jmeridth (approved):
:robot: Re-approving on the current HEAD (8b333a65) after verifying the fix.

The empty-`support-policy` note is fully handled: the check is now `sp != nil && strings.TrimSpace(string(*sp)) != ""` (with the `strings` import), a docstring paragraph explains why, and a new test case (`Release with empty support policy falls through to failure`) asserts an empty field now reaches `Failed` instead of a false `Passed`. Scoping the change to DO-05.01 and leaving the sibling steps on the `!= nil` idiom is a reasonable call. No blockers.

eddie-knight:
We need to have ops add @vinayada1 to the maintainer group so that she can merge her own approved PRs:

https://github.com/orgs/ossf/teams/orbit-pvtr-plugin-maintainers

PR #407, opened 2026-07-28
chore: remove retired OSPS-BR-01.02 support

vinayada1:
## Summary

OSPS-BR-01.02 (branch-name sanitization) was retired from the OSPS Baseline in [ossf/security-baseline#443](https://github.com/ossf/security-baseline/pull/443), where its intent was consolidated into a broadened **BR-01.01** covering all untrusted metadata executed without contributor review. This PR removes the now-retired requirement from the scanner.

## Changes

- **Dispatch map** (`evaluation_plans/evaluation-plans.go`) — removed the `OSPS-BR-01.02` → `CicdBranchNameSanitized` entry.
- **Step code** (`evaluation_plans/osps/build_release/steps.go`) — deleted the now-dead `CicdBranchNameSanitized` step, its `checkWorkflowFileForBranchNameUsage` helper, and the `alwaysUnsafeBranchVars` / `pullRequestOnlyUnsafeBranchVars` regexes. Shared helpers and the BR-01.01 untrusted-inputs check are untouched.
- **Tests** (`evaluation_plans/osps/build_release/steps_test.go`) — removed the two branch-name fixtures and 8 associated tests.
- **Catalogs** — removed the `OSPS-BR-01.02` assessment-requirement block from all three bundled catalogs (`2025_10`, `2026_02`, `backwards-compatibility-protection`).

The catalog entries were removed in lockstep with the map entry because `TestAllCatalogAssessmentIDsHaveSteps` enforces catalog↔step-map sync.

## Verification

- `go build ./...` — clean
- `go test ./...` — all pass
- `go vet ./...` — clean
- `golangci-lint run` (v2.11.4, matches CI) — 0 issues
- No residual references to the removed ID or symbols

## Follow-up

#406 tracks extending BR-01.01 to fully cover the branch-ref variables (`github.base_ref`, `github.event.pull_request.base.ref`, and trigger-aware `github.ref` / `github.ref_name`) that the retired check handled.

Closes #343

Issue #406, opened 2026-07-28
Ensure BR-01.01 fully covers branch-ref variables from retired BR-01.02


vinayada1:
## Background

OSPS-BR-01.02 (branch-name sanitization) was retired upstream in [ossf/security-baseline#443](https://github.com/ossf/security-baseline/pull/443) and removed from this scanner in #343. Per that upstream change, **BR-01.01 was broadened to cover "all untrusted metadata executed without contributor review"** — which explicitly now includes branch names.

However, the scanner's BR-01.01 implementation (`CicdSanitizedInputParameters` / `checkWorkflowFileForUntrustedInputs` in `evaluation_plans/osps/build_release/steps.go`) does **not** yet detect all of the branch-ref variables that the retired BR-01.02 code (`checkWorkflowFileForBranchNameUsage`) used to flag. As a result, BR-01.01 currently has a coverage gap relative to the consolidated requirement.

## Coverage gap

The surviving `untrustedVars` regex already covers `github.head_ref` and `github.event.pull_request.head.ref`, but it is **missing** the following variables that the deleted BR-01.02 code checked:

Always attacker-controllable (safe to add directly to `untrustedVars`):
- `github.base_ref`
- `github.event.pull_request.base.ref`

Only attacker-controllable in `pull_request` / `pull_request_target` workflows (needs trigger-aware handling — see caveat):
- `github.ref`
- `github.ref_name`

## Important caveat

`github.ref` and `github.ref_name` must **not** be added unconditionally. They are benign in `push`-triggered workflows and only attacker-controllable in `pull_request` / `pull_request_target` workflows. The retired BR-01.02 code handled this with a trigger check (`hasPullRequestTrigger`), but the current `checkWorkflowFileForUntrustedInputs` is **not trigger-aware** and flags matches unconditionally. Adding these two variables naively would produce **false positives** on push-triggered workflows.

Porting them correctly therefore means reintroducing the trigger-awareness logic, not just extending the regex.

## Proposed work

- [ ] Add `github.base_ref` and `github.event.pull_request.base.ref` to the untrusted-inputs detection for BR-01.01.
- [ ] Reintroduce trigger-aware handling so `github.ref` / `github.ref_name` are flagged only in `pull_request` / `pull_request_target` workflows (avoiding push-workflow false positives).
- [ ] Add/restore tests covering these variables and the push-vs-PR trigger distinction (the deleted BR-01.02 tests are a useful reference).
[...]

PR #410, opened 2026-07-28
fix(OSPS-BR-01.01): broaden untrusted-input coverage to branch refs

vinayada1:
## Summary

Broadens **OSPS-BR-01.01** (untrusted-input detection) to cover the branch-ref command-injection vectors that were previously handled by the now-retired **OSPS-BR-01.02**, so no coverage is lost when BR-01.02 is removed (see #407).

## Changes

- Add `github.base_ref` and `github.event.pull_request.base.ref` to the unconditional `untrustedVars` set (always attacker-controllable via fork PRs).
- Add trigger-aware handling so `github.ref` and `github.ref_name` are flagged **only** in `pull_request` / `pull_request_target` workflows, avoiding false positives on push-triggered workflows. BR-01.01 owns its own `pullRequestOnlyUntrustedVars` regex so it is unaffected when BR-01.02 is removed.
- Add tests covering the new variables and the push-vs-PR trigger distinction.

## Notes

- Independent of #407 and safe in **any merge order** — landing this first guarantees there is no coverage gap.

Refs #406

### DIFF
--- a/evaluation_plans/osps/build_release/steps.go
+++ b/evaluation_plans/osps/build_release/steps.go
@@ -33,6 +33,12 @@ var (
 		`github\.event\.pull_request\.head\.label|` +
 		`github\.event\.pull_request\.head\.repo\.default_branch|` +
 		`github\.head_ref).*`)
+
+	// Refs that are only attacker-controllable in pull_request /
+	// pull_request_target workflows. On push they are trusted, so they are
+	// checked separately and flagged only for PR-triggered workflows.
+	pullRequestOnlyUntrustedVars = regexp.MustCompile(`.*(github\.ref\b|` +
+		`github\.ref_name).*`)
 )
 
 // checkAllWorkflows fetches the repository's workflow files and evaluates each
@@ -99,14 +105,25 @@ func CicdSanitizedInputParameters(payload data.Payload) (gemara.Result, string, 
 		"GitHub Workflows variables do not contain untrusted inputs")
 }
 
-// checkWorkflowFileForUntrustedInputs checks a workflow for known untrusted
-// GitHub Actions context variables used directly in run: steps.
-// These variables (e.g. issue titles, PR bodies, commit messages) are
-// user-controllable and can lead to command injection when interpolated
-// into shell scripts without sanitization.
+// checkWorkflowFileForUntrustedInputs flags GitHub Actions context variables
+// used directly in run: steps that outside contributors can control (e.g. issue
+// titles, PR bodies, commit messages, branch refs). Interpolating them into a
+// shell script without sanitization can lead to command injection.
+//
+// github.ref and github.ref_name are only attacker-controllable in
+// pull_request / pull_request_target workflows, so they are flagged only there.
 func checkWorkflowFileForUntrustedInputs(workflow *actionlint.Workflow) (bool, string) {
 
 	var message strings.Builder
+
+	// PR triggers make github.ref and github.ref_name attacker-controllable.
+	hasPullRequestTrigger := false
+	for _, event := range workflow.On {
+		if event.EventName() == "pull_request" || event.EventName() == "pull_request_target" {
+			hasPullRequestTrigger = true
+			break
+		}
+	}
 
 	for _, job := range workflow.Jobs {
 		if job == nil {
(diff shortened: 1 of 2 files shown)


vinayada1:
This PR makes **OSPS-BR-01.01** more comprehensive now that **OSPS-BR-01.02** will be retired (#407). It folds BR-01.02's branch-ref command-injection coverage into BR-01.01 — adding `github.base_ref` / `github.event.pull_request.base.ref` unconditionally, and `github.ref` / `github.ref_name` only for pull_request-triggered workflows — so there is no coverage gap in any merge order. Addresses #406.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -32,8 +32,16 @@ var (
 		`github\.event\.pull_request\.head\.ref|` +
 		`github\.event\.pull_request\.head\.label|` +
 		`github\.event\.pull_request\.head\.repo\.default_branch|` +
+		`github\.event\.pull_request\.base\.ref|` +
+		`github\.base_ref|` +
```
jmeridth: :robot: (low, non-blocking) Faithful port, so not a regression — but worth fixing while you're here: `github.base_ref`/`pull_request.base.ref` aren't "always attacker-controllable." They're the PR *target* branch (an existing upstream branch a maintainer created), so a fork PR can't inject metacharacters through them — GitHub's docs and the Security Lab source cited above both list only `head_ref`. Since a match returns a hard `Failed`, a benign `git diff origin/${{ github.base_ref }}...HEAD` (diff-based CI) false-fails BR-01.01 with nothing to sanitize. Either drop `base_ref` from the unconditional set, or keep it and reword the comment/test to "flagged as defense-in-depth; the base ref is maintainer-controlled, not attacker-injectable."

jmeridth (dismissed):
:robot: Approving — multi-model review (Opus 4.8, Sonnet 5, Fable 5). First review on this PR.

Correct and well-tested, verified across all three models plus direct regex checks. The `github\.ref\b` boundary matches `github.ref`/`github.ref_name` while correctly rejecting `ref_type`/`ref_protected`; `base_ref` and both `pull_request.*.ref` forms are unconditional; `github.ref`/`ref_name` are excluded from `untrustedVars` and only flagged under `pull_request`/`pull_request_target`. Trigger loop is byte-identical to the retired BR-01.02 check and handles list + map `on:` forms. No regression (alternation only adds matches), tests are faithful and mutation-tested, and it's a clean port. Intentionally not retiring BR-01.02 here (that's #407), so branch-ref vars are transiently double-covered — matching the safe-in-any-merge-order design.

One low, non-blocking note left inline: the "always attacker-controllable" framing for `base_ref`/`pull_request.base.ref` is inaccurate (those are the maintainer-controlled target branch, not attacker-injectable — GitHub's docs and the cited Security Lab source list only `head_ref`), and flagging them returns a hard `Failed` on benign patterns like `git diff origin/${{ github.base_ref }}...HEAD`. It's a faithful port of pre-existing behavior, not a regression, so non-blocking — but a good opportunity to fix since you're refactoring this path. Not blocking. Leaving the merge to you.

evaluation_plans/osps/build_release/steps.go
vinayada1: Good catch — agreed. `base_ref`/`pull_request.base.ref` are the maintainer-controlled target branch, not attacker-injectable (the cited Security Lab source lists only `head_ref`), and hard-failing on `git diff origin/${{ github.base_ref }}...HEAD` is a real false positive. Dropped both from the unconditional `untrustedVars` set and updated the tests/comments accordingly. Pushed.

jmeridth: :robot: (low, non-blocking) Reopening to keep this visible. On the current head, `untrustedVars` no longer lists `base_ref` (good), but the false positive I flagged still ships: `github.base_ref` / `github.event.pull_request.base.ref` remain in `alwaysUnsafeBranchVars`, which `checkWorkflowFileForBranchNameUsage` treats as a hard `Failed` on every trigger (`steps.go:168`). So a benign `git diff origin/${{ github.base_ref }}...HEAD` step still false-fails. The new tests only assert `base_ref` isn't flagged by the *input* check (`checkWorkflowFileForUntrustedInputs`); nothing covers `base_ref` against the branch-name check.

Note this regex is pre-existing (unchanged from the base branch), so it's arguably out of scope for this PR. Your call: drop `base_ref` + `pull_request.base.ref` from `alwaysUnsafeBranchVars` here, or track it separately as a pre-existing issue. Either is fine by me — flagging so it doesn't get lost when this merges.

jmeridth: :robot: (low, non-blocking) Confirmed resolved on the current head (`616253c2`). `base_ref` / `pull_request.base.ref` are no longer flagged anywhere: the branch-name check that hard-failed them was removed with OSPS-BR-01.02 (#407), and this PR folds ref coverage into the input check with the correct model — `head_ref` always flagged, `github.ref`/`ref_name` only on `pull_request` / `pull_request_target`, `base_ref` never. `TestUntrustedInputsBranchRefCoverage` covers all of it and passes. Thanks for the rework. Resolving.

jmeridth (approved):

PR #355, opened 2026-07-03
fix(OSPS-LE-02.01): pass deprecated but OSI/FSF-approved SPDX IDs

joanagmaia:
## Summary

- Fixes false negatives in `OSPS-LE-02.01` for deprecated-but-approved SPDX IDs (`AGPL-3.0`, `GPL-2.0`, `GPL-3.0`, `LGPL-2.1`) that GitHub's License API still returns in deprecated form.
- Drops the `|| license.IsDeprecatedLicenseId` clause from the fail condition in `GoodLicense` — licenses now fail only when neither OSI- nor FSF-approved.
- Deprecated-but-approved IDs are tracked separately and surfaced as a non-failing advisory note in the passing message, so maintainers are informed of the SPDX identifier hygiene issue.

## Root cause

`evaluation_plans/osps/legal/steps.go` line 95 (before fix):

```go
if (!license.IsOsiApproved && !license.IsFsfLibre) || license.IsDeprecatedLicenseId {
    badLicenses = append(badLicenses, spdx_id)
}
```

The `|| license.IsDeprecatedLicenseId` clause caused any deprecated SPDX ID to be treated as a bad license regardless of approval status, producing the misleading failure message "These licenses are not OSI or FSF approved: AGPL-3.0" for projects like Grafana and Renovate.

## Test plan

- [ ] Existing 5 `TestGoodLicense` cases unchanged and passing
- [ ] New case: `AGPL-3.0` (`isOsiApproved:true, isFsfLibre:true, isDeprecatedLicenseId:true`) → `Passed` with deprecation note
- [ ] New case: `MIT AND AGPL-3.0` mix → `Passed` with note listing only `AGPL-3.0`
- [ ] New case: `AGPL-3.0 AND BadLicense` mix → `Failed` referencing only `BadLicense`
- [ ] New case: deprecated + unapproved → `Failed`
- [ ] `make vet && make test` green

Closes #347

> Note: once merged and released, linuxfoundation/insights#1898 and linuxfoundation/insights#1735 should be re-evaluated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #419, opened 2026-08-03
fix(OSPS-LE-02.01): handle SPDX WITH exceptions, + suffix, and grouping parens

### LINKED ISSUE #234: OSPS-LE-02.01 incorrectly fails for licenses with SPDX WITH exceptions
## Problem

Projects with valid licenses that include SPDX WITH exception clauses (e.g., Apache-2.0 WITH LLVM-exception) are incorrectly failing the OSPS-LE-02.01 evaluation. The license expression is not being parsed correctly — the `WITH <exception>` suffix is treated as part of the license ID, causing it to fail validation against the SPDX license list.

Additionally, `NOASSERTION` — a valid SPDX keyword — is being flagged as an invalid license instead of being skipped.

### Example

The https://github.com/kokkos/kokkos?tab=License-1-ov-file has a valid Apache-2.0 license but is reported as failing OSPS-LE-02.01:

- [Kokkos Project](https://insights.linuxfoundation.org/project/kokkos-project/repository/kokkos-kokkos/security?timeRange=past365days&start=2025-02-04&end=2026-02-04)

## Expected Behavior
- Licenses with WITH exception clauses should be recognized by their base license ID (e.g., Apache-2.0 WITH LLVM-exception → Apache-2.0).
- NOASSERTION should be treated as a valid value and not flagged as an unrecognized license.

### LINKED ISSUE #1928: License report is wrong for Kokkos project
License report is wrong for Kokkos project

Blocked by pvtr-github repo Issue: https://github.com/revanite-io/pvtr-github-repo/issues/234Connect your Github account 

License Status of Kokkos Repository
The Kokkos repository is licensed under Apache License v2.0 with LLVM Exceptions [kokkos/LICENSE at develop · kokkos/kokkos +2](https://github.com/kokkos/kokkos/blob/develop/LICENSE), which is an OSI-approved open source license.

Why the Report is Wrong
Apache 2.0 is OSI and FSF approved: The Apache Software Foundation and the Free Software Foundation agree that the Apache License 2.0 is a free-software license, compatible with the GNU General Public License (GPL) version 3 [Wikipedia](https://en.wikipedia.org/wiki/Apache_License).

LICENSE file exists: The repository clearly has a LICENSE file in both the develop and master branches containing the full Apache 2.0 license text with LLVM exceptions.

"NOASSERTION" finding is incorrect: The report claims the license is "NOASSERTION," which typically means a license scanning tool couldn't identify the license. This appears to be an error in whatever tool generated this report, as the repository clearly displays its Apache 2.0 with LLVM Exceptions license.

Conclusion
The Kokkos repository already meets requirement OSPS-LE-02.01. It has:

✅ A LICENSE file in the repository

✅ An OSI-approved license (Apache 2.0)

✅ An FSF-approved free software license (Apache 2.0)

The LLVM exception is an addition that provides GPLv2 compatibility and compiler-output exceptions, but doesn't change the fact that the base Apache 2.0 license is both OSI and FSF approved.

joanagmaia:
## Problem

Projects with valid SPDX licenses that include `WITH <exception>` clauses (e.g. `Apache-2.0 WITH LLVM-exception`) were incorrectly failing `OSPS-LE-02.01`. The `splitSpdxExpression` tokenizer only split on ` AND ` / ` OR `, so exception-bearing expressions were passed verbatim to the SPDX license list lookup — which naturally found no exact match and flagged them as unapproved.

The same tokenizer also failed silently on:
- Grouped expressions with parentheses, e.g. `(MIT OR Apache-2.0)`
- Deprecated `+` (or-later) suffix, e.g. `GPL-2.0+`

Closes #234
Helps close linuxfoundation/insights#1928

## Changes

`splitSpdxExpression` (`evaluation_plans/osps/legal/steps.go`) now normalizes expressions before splitting:

1. Strips grouping `(` / `)` (no semantic effect on approval)
2. Strips ` WITH <exception>` suffix per token — exception clauses don't affect OSI/FSF approval of the base license
3. Strips deprecated `+` (or-later) suffix per token

`NOASSERTION` handling was already in place and is unchanged.

## Tests

Four new cases added to `TestSplitSpdxExpression` and one to `TestGoodLicense` (Kokkos-pattern `Apache-2.0 WITH LLVM-exception` → `Passed`). All existing tests continue to pass.

## Local verification against kokkos/kokkos

Running the scanner against `kokkos/kokkos` (the repo cited in #234) returns `Needs Review` rather than `Passed`. This is expected and unrelated to this fix.

GitHub uses [licensee](https://github.com/licensee/licensee) to detect licenses and returns `spdx_id: NOASSERTION` for kokkos because the `LICENSE` file is Apache-2.0 text with LLVM exception clauses appended — the modified text doesn't match licensee's known templates. The scanner falls through to the `NeedsReview` path (license file present, identity unknown) regardless of this fix.

This fix applies when the SPDX expression reaches the scanner as a string — either via GitHub's API (for repos whose license text licensee *can* classify) or via a Security Insights file that explicitly declares `license.expression: Apache-2.0 WITH LLVM-exception`. For kokkos to reach `Passed`, the kokkos maintainers would need to add a Security Insights file with the correct expression.

## Background

Initial fix attempt for the same underlying issue was started in #355. This PR supersedes it with a broader fix covering the full class of SPDX expression constructs that break the tokenizer.

### DIFF
--- a/evaluation_plans/osps/legal/steps.go
+++ b/evaluation_plans/osps/legal/steps.go
@@ -42,10 +42,23 @@ func getLicenseList(payload data.Payload, makeApiCall func(string, bool) ([]byte
 }
 
 func splitSpdxExpression(expression string) (spdx_ids []string) {
+	// Remove grouping parentheses; they affect evaluation order, not approval.
+	expression = strings.ReplaceAll(expression, "(", " ")
+	expression = strings.ReplaceAll(expression, ")", " ")
 	a := strings.Split(expression, " AND ")
 	for _, aa := range a {
 		b := strings.Split(aa, " OR ")
-		spdx_ids = append(spdx_ids, b...)
+		for _, token := range b {
+			token = strings.TrimSpace(token)
+			// Strip " WITH <exception>" — the exception clause is irrelevant to
+			// OSI/FSF approval; only the base license ID matters.
+			if idx := strings.Index(token, " WITH "); idx != -1 {
+				token = token[:idx]
+			}
+			// Strip deprecated "+" (or-later) suffix so "GPL-2.0+" resolves to "GPL-2.0".
+			token = strings.TrimSuffix(token, "+")
+			spdx_ids = append(spdx_ids, token)
+		}
 	}
 	return
 }
--- a/evaluation_plans/osps/legal/steps.go
+++ b/evaluation_plans/osps/legal/steps.go
@@ -55,7 +55,8 @@ func splitSpdxExpression(expression string) (spdx_ids []string) {
 			if idx := strings.Index(token, " WITH "); idx != -1 {
 				token = token[:idx]
 			}
-			// Strip deprecated "+" (or-later) suffix so "GPL-2.0+" resolves to "GPL-2.0".
+			// Strip the "+" (or-later) operator so any "id+" form (e.g. "Apache-2.0+")
+			// resolves to its base license ID.
 			token = strings.TrimSuffix(token, "+")
 			spdx_ids = append(spdx_ids, token)
 		}
--- a/evaluation_plans/osps/legal/steps_test.go
+++ b/evaluation_plans/osps/legal/steps_test.go
@@ -273,6 +273,26 @@ func TestSplitSpdxExpression(t *testing.T) {
 			input:    "",
 			expected: []string{""},
 		},
+		{
+			name:     "WITH exception suffix",
+			input:    "Apache-2.0 WITH LLVM-exception",
+			expected: []string{"Apache-2.0"},
+		},
+		{
+			name:     "WITH exception inside grouped OR",
+			input:    "(GPL-2.0 WITH Classpath-exception-2.0) OR MIT",
+			expected: []string{"GPL-2.0", "MIT"},
+		},
+		{
+			name:     "Deprecated + suffix",
+			input:    "GPL-2.0+",
+			expected: []string{"GPL-2.0"},
+		},
+		{
+			name:     "Parentheses grouping",
+			input:    "(MIT OR Apache-2.0)",
+			expected: []string{"MIT", "Apache-2.0"},
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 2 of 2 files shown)


evaluation_plans/osps/legal/steps.go
```diff
@@ -42,10 +42,23 @@ func getLicenseList(payload data.Payload, makeApiCall func(string, bool) ([]byte
...
 		b := strings.Split(aa, " OR ")
-		spdx_ids = append(spdx_ids, b...)
+		for _, token := range b {
+			token = strings.TrimSpace(token)
+			// Strip " WITH <exception>" — the exception clause is irrelevant to
+			// OSI/FSF approval; only the base license ID matters.
+			if idx := strings.Index(token, " WITH "); idx != -1 {
+				token = token[:idx]
+			}
+			// Strip deprecated "+" (or-later) suffix so "GPL-2.0+" resolves to "GPL-2.0".
```
jmeridth: :robot: (low, non-blocking) Small accuracy note: `GPL-2.0+` and the five other `+`-suffixed GPL/LGPL IDs exist verbatim as deprecated entries in the SPDX license list, so that example already resolved before this change. The strip is still worthwhile because the SPDX grammar allows `+` on any license ID (e.g. `Apache-2.0+`, which has no list entry). One side effect: the deprecation notice and trace logs now report the stripped ID (`GPL-2.0`) rather than the declared one (`GPL-2.0+`). No pass/fail change on current SPDX data (checked all six).

```suggestion
			// Strip the "+" (or-later) operator so any "id+" form (e.g. "Apache-2.0+")
			// resolves to its base license ID.
```

jmeridth (dismissed):
:robot: (informational, non-blocking) Multi-model AI-assisted review (Claude Opus, Sonnet, and Haiku agents, findings cross-verified). Independently verified: build, go vet, and the full test suite pass on the PR branch, and the tokenizer was probed against grouped, malformed, `+`, and `WITH` edge cases plus the live SPDX license list. No correctness issues found. One non-blocking accuracy nit inline. Bonus: `(MIT)AND(GPL-2.0)` without operator spacing now tokenizes correctly too, since the paren replacement introduces the needed spaces.

evaluation_plans/osps/legal/steps.go
joanagmaia: Good catch — updated the comment to reflect that. Thanks!

eddie-knight (approved):

PR #420, opened 2026-08-03
feat: implement OSPS-QA-02.02 SBOM check for released assets

### LINKED ISSUE #342: Implement OSPS-QA-02.02
OSPS-QA-02.02 is still a NotImplemented stub. The older control-level issue was closed after QA-02.01 work, but QA-02.02 was missed.

Work needed:

- Replace the stub with a real check
- Add focused tests
- Link back to ossf/pvtr-github-repo-scanner#24 for history

vinayada1:
## Summary

Implements **OSPS-QA-02.02**, replacing the previous `reusable_steps.NotImplemented` stub with a real assessment step, `quality.ReleasesHaveSBOM`.

> Control: *When the project has made a release, all compiled released software assets MUST be delivered with a software bill of materials (SBOM).* (Maturity Level 3)

Closes #342. Related history: #24.

## How it works

Security Insights has no SBOM field, so the only observable evidence is the set of assets attached to GitHub releases (REST). GitHub's auto-generated source archives are not part of the REST asset list, so every asset seen is a deliberately published artifact.

Each release asset is classified into one of three buckets:
- **SBOM documents** — CycloneDX / SPDX / CDX markers (signatures, checksums, and binaries that merely embed an SBOM marker are excluded).
- **Definitely-compiled artifacts** — by binary extension (`.exe`, `.so`, `.dll`, `.dylib`, `.a`, …).
- **Ambiguous artifacts** — archives (`.tar.gz`, `.zip`, …) and extensionless files, which commonly bundle native binaries and are routed to review.

## Why this check returns `NeedsReview` instead of `Failed`

**A missing *public* SBOM is not evidence of non-compliance, so this control intentionally never returns `Failed`.**

Under the EU Cyber Resilience Act (Regulation (EU) 2024/2847), a manufacturer must *create* an SBOM as part of the product's technical documentation and provide it to market-surveillance authorities on reasoned request — but **publishing the SBOM to users is optional** ("If the manufacturer decides to make available the SBOM…"). An SBOM can therefore legitimately be retained privately, delivered through a separate distribution channel, or produced only for audit.

In addition, Privateer runs this scan from a **software-publisher** perspective — the scanner observes what is publicly attached to releases and cannot see privately-retained SBOMs, and the operator is often not the party that produces the SBOM.

Because the absence of a public SBOM is genuinely ambiguous (compliant-but-private vs. actually-missing), the check surfaces it as `NeedsReview` with a message pointing the reviewer to publisher evidence, rather than asserting a failure it cannot substantiate. This rationale is also documented in the doc comment on `ReleasesHaveSBOM`.

## Result matrix

| Situation | Result | Confidence |
|---|---|---|
| No published (non-draft) releases | `NotApplicable` | High |
| Releases exist but no attached assets | `NotApplicable` | Low |
| Compiled/archived assets present, all releases have an SBOM | `Passed` | Medium/Low |
| Only an SBOM present, no compiled/archived assets | `Passed` | Low |
| Compiled/archived assets present, some release missing an SBOM | `NeedsReview` | Low |
| No compiled/archived assets and no SBOM | `NeedsReview` | Low |
| Release data unavailable or retrieval error | `NeedsReview` | Low |

## Data-layer changes (`data/rest-data.go`)

- Added release **pagination** (`per_page=100`, loops until a short page).
- Added a `Draft` field to `ReleaseData`; draft releases are excluded from both the published-release precondition and the asset scan.
- Release-fetch errors are now captured in `RestData.ReleasesError` and propagated so retrieval failures become `NeedsReview` rather than a false `NotApplicable`.

## Tests

- `TestReleasesHaveSBOM` — result matrix incl. drafts, pagination, retrieval errors, ambiguous/archived/extensionless assets.
- `TestIsSBOMAsset`, `TestIsCompiledReleaseAsset`, `TestIsAmbiguousBinaryAsset` — classifier unit tests incl. sidecar/false-positive guards.
- `TestGetReleasesPaginatesAndDecodesDrafts`, `TestGetReleasesReturnsDecodeError` — data-layer pagination/draft/error handling.

`go build ./...`, `go vet ./...`, and `go test ./...` (incl. `-race`) all pass.

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -208,6 +208,7 @@ func (r *RestData) checkFileInSubdir(dir, filename string) string {
 	}
 	return ""
 }
+
 // dependency-update tools (Dependabot, Renovate). Their presence is direct
 // evidence a repository manages its dependencies, observable even when
 // security-insights.yml is absent.
@@ -501,10 +502,12 @@ func (r *RestData) getReleases() error {
 		endpoint := fmt.Sprintf("%s/repos/%s/%s/releases?per_page=%d&page=%d", APIBase, r.owner, r.repo, perPage, page)
 		responseData, err := r.MakeApiCall(endpoint, true)
 		if err != nil {
+			r.Releases = nil
 			return fmt.Errorf("failed to fetch releases page %d: %w", page, err)
 		}
 		var releases []ReleaseData
 		if err := json.Unmarshal(responseData, &releases); err != nil {
+			r.Releases = nil
 			return fmt.Errorf("failed to decode releases page %d: %w", page, err)
 		}
 		r.Releases = append(r.Releases, releases...)
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -40,6 +40,7 @@ type RestData struct {
 	PrivateVulnReporting        PrivateVulnReporting
 	SecurityPolicy              SecurityPolicy
 	Releases                    []ReleaseData
+	ReleasesError               error `json:"-" yaml:"-"`
 	contents                    RepoContent
 	ghClient                    *github.Client `json:"-" yaml:"-"`
 	HttpClient                  HttpClient     `json:"-" yaml:"-"`
@@ -55,6 +56,7 @@ type ReleaseData struct {
 	Name    string         `json:"name"`
 	TagName string         `json:"tag_name"`
 	URL     string         `json:"url"`
+	Draft   bool           `json:"draft"`
 	Assets  []ReleaseAsset `json:"assets"`
 }
 
@@ -89,7 +91,7 @@ func (r *RestData) Setup() error {
 		_ = r.getWorkflowPermissions()
 	})
 	wg.Go(func() {
-		_ = r.getReleases()
+		r.ReleasesError = r.getReleases()
 	})
 	wg.Go(func() {
 		r.getPrivateVulnReporting()
@@ -493,12 +495,23 @@ func (r *RestData) rootHasDir(name string) bool {
 }
 
 func (r *RestData) getReleases() error {
-	endpoint := fmt.Sprintf("%s/repos/%s/%s/releases", APIBase, r.owner, r.repo)
-	responseData, err := r.MakeApiCall(endpoint, true)
-	if err != nil {
-		return err
+	const perPage = 100
+	r.Releases = nil
+	for page := 1; ; page++ {
+		endpoint := fmt.Sprintf("%s/repos/%s/%s/releases?per_page=%d&page=%d", APIBase, r.owner, r.repo, perPage, page)
+		responseData, err := r.MakeApiCall(endpoint, true)
+		if err != nil {
+			return fmt.Errorf("failed to fetch releases page %d: %w", page, err)
+		}
+		var releases []ReleaseData
+		if err := json.Unmarshal(responseData, &releases); err != nil {
+			return fmt.Errorf("failed to decode releases page %d: %w", page, err)
+		}
+		r.Releases = append(r.Releases, releases...)
+		if len(releases) < perPage {
+			return nil
+		}
 	}
-	return json.Unmarshal(responseData, &r.Releases)
 }
 
 func (r *RestData) getWorkflowPermissions() error {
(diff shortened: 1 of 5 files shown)


evaluation_plans/osps/quality/steps.go
```diff
@@ -299,6 +299,312 @@ func DocumentsTestMaintenancePolicy(payload data.Payload) (result gemara.Result,
...
+	for _, suffix := range signatureOrChecksumSuffixes {
+		if strings.HasSuffix(lower, suffix) {
+			return true
+		}
+	}
+	for _, token := range sbomTokens(lower) {
+		if token == "checksums" || token == "checksum" {
+			return true
+		}
+	}
```
jmeridth, line 606: :robot: (low, non-blocking) Extensionless checksum manifests like `SHA256SUMS`, `MD5SUMS`, or `terraform_1.9.0_SHA256SUMS` tokenize to `sha256sums`/`md5sums` and match neither the dotted suffixes nor the `checksum(s)` tokens, so they fall through to `isAmbiguousBinaryAsset` and count as plausibly-compiled artifacts. Verified this never flips the Passed/NeedsReview verdict, but the NeedsReview message then claims the release is "publishing compiled or archived software" when it only ships a checksum listing.
```suggestion
	for _, token := range sbomTokens(lower) {
		switch token {
		case "checksums", "checksum",
			"shasums", "sha1sums", "sha256sums", "sha512sums", "md5sums", "b2sums":
			return true
		}
	}
```

data/rest-data.go
```diff
@@ -493,12 +495,23 @@ func (r *RestData) rootHasDir(name string) bool {
...
+		endpoint := fmt.Sprintf("%s/repos/%s/%s/releases?per_page=%d&page=%d", APIBase, r.owner, r.repo, perPage, page)
+		responseData, err := r.MakeApiCall(endpoint, true)
+		if err != nil {
+			return fmt.Errorf("failed to fetch releases page %d: %w", page, err)
+		}
+		var releases []ReleaseData
+		if err := json.Unmarshal(responseData, &releases); err != nil {
+			return fmt.Errorf("failed to decode releases page %d: %w", page, err)
+		}
+		r.Releases = append(r.Releases, releases...)
```
jmeridth, line 513: :robot: (low, non-blocking) If a later page fails after earlier pages succeed, `r.Releases` keeps the partial list while `ReleasesError` is set. `ReleasesHaveSBOM` guards on `ReleasesError` first so it's safe, but the other `payload.Releases` consumers in `build_release`, `docs`, and `legal` don't check it and now see a truncated list where the old code gave them nil. Rare path and partial data is arguably still useful, so just flagging the semantics change — resetting `Releases = nil` on error would restore the old all-or-nothing behavior if you'd rather keep it.

jmeridth (commented):
:robot: (low, non-blocking) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass), every finding verified by executing the classifiers or tracing code paths.

No blockers found. `go build`, `go vet`, `go test ./...`, and `go test -race` all pass on the PR branch. Two low, non-blocking findings posted as line comments: the `SHA256SUMS`-style checksum-manifest classification gap (never flips a verdict, only the NeedsReview message text) and the partial `Releases` list left behind on mid-pagination errors (only affects consumers that don't check `ReleasesError`).

Verified clean: concurrency around `ReleasesError` (race detector clean), no recursion in the classifier call graph, draft assets excluded from `totalAssets`, nil `RestData` guard, exact-100 pagination boundary, the `NotImplemented` swap has no side effects, and the doc comments and PR result matrix match the implemented behavior.

jmeridth (dismissed):
:robot: (low, non-blocking) Approving — multi-model verification found no correctness blockers; the two line comments are non-blocking message-fidelity notes.

evaluation_plans/osps/quality/steps.go
vinayada1: Fixed in 21a5eb1: extended `isSignatureOrChecksumAsset` to match `*SUMS` tokens (sha256sums, md5sums, etc.), with a regression test in `TestIsSignatureOrChecksumAsset`.

data/rest-data.go
vinayada1: Fixed in 21a5eb1: `getReleases` now resets `Releases = nil` on any page error, restoring all-or-nothing semantics; covered by `TestGetReleasesResetsPartialListOnLaterPageError`.

jmeridth (approved):
:robot: (low, non-blocking) Re-approving at 21a5eb1 — verified both review comments were addressed exactly as suggested (checksum-manifest `*SUMS` tokens now classify as checksum companions; `getReleases` resets to nil on page errors, restoring all-or-nothing semantics), with regression tests covering each. Full data and quality suites pass locally.

PR #421, opened 2026-08-03
feat(OSPS-VM-06.02): enforce SAST checks on code changes

### LINKED ISSUE #416: Implement OSPS-VM-06.02
The scanner does not yet evaluate this OSPS Baseline 2026-02-19 requirement (currently returns "Not Run").

Requirement:
> While active, all changes to the project's codebase MUST be automatically evaluated against a documented policy for security weaknesses and blocked in the event of violations except when declared and suppressed as non-exploitable.

Maturity: Level 3

Work needed:
- Confirm current requirement text/applicability against OSPS Baseline 2026-02-19
- Implement detection of an automated SAST gate/status check enforced on changes
- Add unit tests

vinayada1:
## Summary

Implement OSPS-VM-06.02 by verifying that every code change is evaluated by a documented SAST policy and that the corresponding check is required on the default branch.

- detect SAST CI integrations declared in Security Insights
- inspect pull request and push workflows, including local reusable workflows
- correlate documented policy evidence with required GitHub Actions check contexts for the same SAST tool
- conservatively return `NeedsReview` when workflow coverage or branch protection cannot be proven
- prevent unrelated tools, ambiguous check names, or unsupported branch globs from producing false passes

## Validation

- `gofmt`
- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `golangci-lint run ./...` (0 issues)
- `go mod tidy` produces no changes

Closes #416

### DIFF
--- a/evaluation_plans/osps/vuln_management/steps.go
+++ b/evaluation_plans/osps/vuln_management/steps.go
@@ -3,11 +3,12 @@ package vuln_management
 import (
 	"fmt"
 	"regexp"
-	"slices"
 	"strings"
 
 	"github.com/gemaraproj/go-gemara"
 	"github.com/ossf/pvtr-github-repo-scanner/data"
+	"github.com/ossf/si-tooling/v2/si"
+	"github.com/rhysd/actionlint"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -73,19 +74,473 @@ func HasSecContact(payload data.Payload) (result gemara.Result, message string,
...
+				if id := matchSast(exec.Run.Value); id != "" {
+					return id
+				}
+			}
+		}
+		if step.Name != nil {
+			if id := matchSast(step.Name.Value); id != "" {
+				return id
+			}
+		}
```
jmeridth: :robot: (high) `matchSast` substring-matches over `run:` text and step names, so a *mention* of a tool is indistinguishable from an *invocation*. Reproduced with a test: a workflow whose only step is `run: echo "TODO wire up CodeQL"`, plus a Security Insights entry documenting CodeQL with a ruleset, plus a required check named after the job, returns `Passed` with High confidence — no SAST ever executes. That contradicts the PR's stated goal of preventing false passes.

Dropping the `step.Name` match alone won't fix it — the `run:` text channel reproduces the same result. One direction: only accept `uses:` action refs as definite evidence, and for `run:` commands require the identifier to appear as a command token (start of the line or after `&&`/`|`/`;`) rather than anywhere in the text.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -73,19 +74,473 @@ func HasSecContact(payload data.Payload) (result gemara.Result, message string,
...
+	"sonarcloud",
+	"sonarqube",
+	"sonar-scanner",
+	"sonarsource",
+	"checkmarx",
+	"snyk code",
+	"bandit",
+	"gosec",
+	"brakeman",
+	"bearer",
```
jmeridth, line 95: :robot: (medium) `"bearer"` collides with ordinary auth text: `matchSast("curl -H 'Authorization: Bearer $TOKEN'")` returns `"bearer"` (verified by execution), and bearer-token text in workflow steps is ubiquitous. A repo with no SAST at all then gets phantom SAST evidence — a correct `Failed` degrades to `NeedsReview` when branch protection isn't observable, or produces a `Failed` with a wrong message when it is. Scoping the identifier to actual Bearer-CLI invocations avoids the collision:
```suggestion
	"bearer/bearer", "bearer scan",
```

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -73,19 +74,473 @@ func HasSecContact(payload data.Payload) (result gemara.Result, message string,
...
+	for i := 0; i < len(pattern); i++ {
+		switch pattern[i] {
+		case '*':
+			if i+1 < len(pattern) && pattern[i+1] == '*' {
+				expression.WriteString(".*")
+				i++
+			} else {
+				expression.WriteString("[^/]*")
+			}
+		case '?', '+', '[', ']':
```
jmeridth, line 221: :robot: (medium) The bail-out to "no match" for `?`/`+`/`[`/`]` is conservative for include filters but inverted for `branches-ignore`: GitHub's `mai+n` (quantifier semantics) matches `main`, so a workflow with `branches-ignore: [mai+n]` never runs for the default branch — yet `branchFilterExcludes` treats the pattern as not excluding and `workflowCoversChanges` claims proven coverage (verified by execution; two independent review passes hit the same path via negated patterns too). The comment above says this avoids "risking a false coverage claim", which doesn't hold in the exclude direction. Suggest a tri-state or direction-aware bail-out: unsupported pattern in an include list → not covered (current), in an ignore/negated position → treat as possibly excluded (uncertain).

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -73,19 +74,473 @@ func HasSecContact(payload data.Payload) (result gemara.Result, message string,
...
+}
+
+// SastEnforcedOnChanges implements OSPS-VM-06.02: all changes to the codebase
+// must be automatically evaluated for security weaknesses and blocked on
+// violations. It confirms both that a SAST tool runs on changes (in CI, per
+// Security Insights or a workflow triggered by pull_request/push) and that it is
+// enforced as a required status check that blocks merges to the default branch.
+func SastEnforcedOnChanges(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	var detection sastDetection
+	if payload.RestData != nil && payload.Insights.Repository != nil {
```
jmeridth, line 584: :robot: (medium) On main this plan included `reusable_steps.HasSecurityInsightsFile`, which surfaced `payload.InsightsError` as `NeedsReview` ("error encountered while parsing Security Insights content"). The new step never consults `InsightsError`, so a repo with a broken `SECURITY-INSIGHTS.yml` that genuinely documents a SAST tool is treated identically to one with no file, and can land on `Failed` with no hint that the file is the problem. Checking `payload.InsightsError` here — either returning `NeedsReview` as before or setting `inspectionBlocked` — restores the lost signal.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -73,19 +74,473 @@ func HasSecContact(payload data.Payload) (result gemara.Result, message string,
...
+			}
+			sources, callBlocked := inspectJob(calledJob, visiting)
+			blocked = blocked || callBlocked
+			if len(sources) > 0 {
+				if calledJob.Name != nil && calledJob.Name.Value != "" {
+					sources = append(sources, calledJob.Name.Value)
+				}
+				if calledJob.ID != nil && calledJob.ID.Value != "" {
+					sources = append(sources, calledJob.ID.Value)
+				}
```
jmeridth, line 440: :robot: (low, non-blocking) The caller side prefers `name:` over the job ID for check contexts, but here both `calledJob.Name` and `calledJob.ID` are appended, so a named called job produces a fabricated alias like `SAST gate / analyze` that GitHub never publishes (it uses name-or-id, name preferred, same as any job). Exact-matching later means this only risks a spurious match against a coincidentally named required check, so non-blocking — but mirroring the caller's name-else-id preference would drop the dead alias.

jmeridth (changes requested):
:robot: (high) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); every reported finding was reproduced by executing the real code, and the headline exploits were independently re-verified before posting.

The structure here is good — separating `evaluateSastEnforcement` from data access made the decision matrix genuinely testable, and most of the conservative machinery (recursion guards, reusable-workflow path resolution, nil-safety, draft/uncertainty handling) checks out cleanly, including under `-race`. But one high finding blocks: free-text substring matching lets a workflow that merely *mentions* a SAST tool (`run: echo "TODO wire up CodeQL"`) produce a High-confidence `Passed` when combined with a documented Insights entry and a matching required check, which contradicts the PR's no-false-passes goal. Details and fix directions are in the five line comments (1 high, 3 medium, 1 low non-blocking).

Also verified but intentionally not blocking, worth knowing: `push` with `branches: [main]` counts as uncertain rather than covered; matrix-expanded check contexts like `Analyze (go)` — CodeQL's normal shape — never exact-match, so real CodeQL repos will tend to land on `NeedsReview` rather than `Passed`; 3-deep reusable-workflow chains flatten check names; and the exported `SastEnforcedOnChanges` has no direct test (all coverage targets unexported helpers). A claimed bug in the only-negative branch-filter case was checked and refuted — GitHub requires at least one positive pattern, so the current behavior is correct there.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed in 5d15791: `run:` scripts are now matched only at a command-token boundary (`matchSastCommand`), `uses:` refs stay definite, and step names are no longer used as evidence — a mention like `echo "TODO wire up CodeQL"` no longer counts as an invocation.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed in 5d15791: bare `bearer` is no longer matched in `run:` text; command matching requires `bearer scan`, so `Authorization: Bearer` headers no longer produce phantom SAST evidence.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed in 5d15791: `matchesGitHubGlob` now returns a `supported` flag and branch filters are direction-aware — an unsupported pattern marks coverage uncertain (NeedsReview) instead of falsely proving inclusion or non-exclusion.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed in 5d15791: `SastEnforcedOnChanges` now sets `inspectionBlocked` when `payload.InsightsError` is set, restoring the parse-error signal (NeedsReview) lost from the main plan.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed in 5d15791: called-job aliasing now mirrors the caller's name-else-id preference instead of appending both, dropping the fabricated `name / id` alias.

jmeridth (approved):
:robot: (low, non-blocking) Approving at 5d15791 — re-ran every original exploit against the new head: bearer auth text and mention-only workflows no longer produce SAST evidence (the false-Pass chain now lands on NeedsReview), unsupported glob patterns are direction-aware and mark coverage uncertain, InsightsError is surfaced again, and called-job aliases follow name-else-id. Legitimate detection (uses: refs, wrapped/pathed commands like `sudo /usr/local/bin/gosec`, `pip install bandit && bandit`) still works, and the full suite passes with -race.

One known tradeoff worth a line in the docs someday, not blocking: command-token matching no longer catches SAST run via `docker run <image> semgrep ...` in a run: script (the common `container:` + `semgrep ci` pattern still detects fine).

PR #422, opened 2026-08-03
feat(OSPS-VM-04): implement published-vulnerability-data and VEX checks

vinayada1:
## Summary

Implements **OSPS-VM-04** — "The project MUST publicly publish data about
discovered vulnerabilities." Both assessment requirements were previously
stubbed as `NotImplemented`:

- **VM-04.01** — publicly publish data about discovered vulnerabilities.
- **VM-04.02** — account for non-affecting vulnerabilities in a VEX document.

## What the checks do

**VM-04.01 (`PublishesVulnerabilityData`)**
- Fetches the repository's **published GitHub Security Advisories** via the REST
  API (`GET /repos/{owner}/{repo}/security-advisories?state=published`).
- ≥1 published advisory → **Passed** (direct public evidence of disclosed
  vulnerabilities).
- Otherwise → **NeedsReview** (never `Failed`): absence of advisories is not a
  real failure — many projects disclose via CVE programs, mailing lists, or
  their own sites rather than GitHub's repository advisory feature. Distinct
  messages cover "signal unobservable" and "a disclosure process exists but
  nothing is published here."

**VM-04.02 (`HasVexDocument`)**
- Scans the **already-fetched repository tree** (no extra API calls) for VEX
  documents — OpenVEX / CSAF / CycloneDX-VEX filenames and `vex/` directories —
  using delimited-token + extension gating to avoid false positives (e.g.
  `convex.json`, `openvex.go` tooling source).
- VEX document found → **Passed**; none → **NeedsReview**.

## Implementation notes

- Signals are gathered in the `data` layer (no per-step network calls); step
  functions stay pure, consistent with existing controls.
- Advisory fetch uses the observed-vs-unknown (`Known`) pattern so an
  unavailable signal degrades to `NeedsReview` rather than a confident result.
- Only the first advisory page is fetched (enough to answer "are there any?");
  a full page sets `CountIsLowerBound` and the pass message reads "at least N".
- VEX detection is a medium-confidence filename/path heuristic; it does not
  parse document content. Generic CSAF filenames do not pass because CSAF
  includes profiles other than VEX.
- The reused GraphQL repository tree currently covers three levels. A VEX file
  nested deeper may be missed and conservatively results in `NeedsReview`.

## Testing

- `go build ./...`, `go vet`, `golangci-lint` (0 issues)
- `go test ./...` — all packages pass, including `-race`
- New unit tests cover advisory parsing (known/unknown/lower-bound), VEX
  detection patterns (incl. tooling/CSAF false positives and nested paths), and
  all step result branches.

## Files changed

- `data/vuln_reporting.go`, `data/rest-data.go`, `data/payload.go`,
  `data/vex.go` (new)
- `evaluation_plans/osps/vuln_management/steps.go`,
  `evaluation_plans/evaluation-plans.go`
- Tests + `README.md` implemented-controls count


data/vex.go
```diff
@@ -0,0 +1,79 @@
...
+// or any structured-data file living under a vex/ directory — while avoiding
+// false positives from words that merely contain "vex".
+func isVexPath(p string) bool {
+	lower := strings.ToLower(p)
+	base := path.Base(lower)
+	dir := path.Dir(lower)
+	ext := path.Ext(base)
+
+	// A structured-data file inside a vex/ directory segment.
+	if structuredDataExtensions[ext] && vexTokenPattern.MatchString("/"+dir+"/") {
```
jmeridth: :robot: (medium) `dir` here is the entire directory path, so a delimited `vex` token in *any* ancestor segment passes: verified by execution, `tools/vex/generate.json`, `.vex/data.json`, and `my-vex-tool/config.json` all return true, which auto-passes OSPS-VM-04.02 for repos whose vex-tooling directories contain unrelated structured files. I realize the vex-directory rule is intentional, but scoping it to the immediate parent keeps every case in the current tests (`vex/bom.json`, `docs/vex/product.json`) while shrinking the false-`Passed` surface:
```suggestion
	if structuredDataExtensions[ext] && vexTokenPattern.MatchString("/"+path.Base(dir)+"/") {
```

data/payload.go
```diff
@@ -123,6 +124,10 @@ func Loader(config *config.Config) (payload any, err error) {
 
 	binaries := analyzeBinaries(tree, treeErr, httpClient, config, owner, repo, graphql.Repository.DefaultBranchRef.Name)
 
+	// Reuse the tree fetched for binary analysis (nil on treeErr) to look for VEX
+	// documents without spending another API call.
+	vexDocuments := detectVexDocuments(tree)
```
jmeridth: :robot: (low, non-blocking) Two related gaps here. First, "nil on treeErr" overclaims: `fetchGraphqlRepoTree` passes a named return straight to `client.Query`, and githubv4 decodes partial data before returning GraphQL-level errors, so on a field-permission error the tree can be non-nil and incomplete while `treeErr` is set. Second, unlike `analyzeBinaries` (which guards `treeErr` and carries `BinaryAnalysis.Err`), nothing threads the error to `HasVexDocument`, so a failed or partial tree fetch yields "No VEX document was found in the repository" — asserting a scan that didn't happen. The result is `NeedsReview` either way, so no verdict flips; flagging because it breaks the observed-vs-unknown pattern this PR establishes nicely for `SecurityAdvisories.Known`. Threading `treeErr` (or `Binaries.Err`) into the step would let the message distinguish "not observed" from "observed none".

data/vex.go
```diff
@@ -0,0 +1,79 @@
...
+		return true
+	}
+
+	// CycloneDX VEX naming, e.g. bom.cdx-vex.json or app.cyclonedx-vex.json, and
+	// OpenVEX naming, e.g. openvex.json or product.openvex.json. Gated on a
+	// structured-data extension so VEX *tooling* source/docs (openvex.go,
+	// cyclonedx-vex.md) are not mistaken for a published VEX document.
+	if structuredDataExtensions[ext] &&
+		(strings.Contains(base, "cdx-vex") || strings.Contains(base, "cyclonedx-vex") || strings.Contains(base, "openvex")) {
+		return true
```
jmeridth: :robot: (low, non-blocking) This branch uses raw `strings.Contains` while everything else in the file is token-delimited, so `isVexPath("openvexation.json")` returns true (verified by execution). Contrived filename, but it contradicts the file's stated design of avoiding matches on words that merely contain the marker. Bounding these markers the same way as the `vex` token (e.g. requiring a delimiter or string boundary around `openvex`/`cdx-vex`/`cyclonedx-vex`) would close it.

jmeridth (commented):
:robot: (medium) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); all reported findings re-verified by executing the real code, including live REST-endpoint checks against real repos.

Cleanest of the recent batch — the advisory fetch mirrors the `PrivateVulnReporting` observed-vs-unknown pattern line-for-line, error mapping and the `CountIsLowerBound` boundary are correct, concurrency is race-clean, and 90+ adversarial executions of the parsers and step branches found no failures. Three line comments posted: one medium that needs an author decision (the ancestor-directory `vex` token match lets `tools/vex/generate.json`-style files auto-pass VM-04.02) and two low non-blocking notes (tree-fetch errors produce a "no VEX found" message without a `Known`-style guard; the OpenVEX branch uses substring rather than token matching).

Pre-existing quirks noted for awareness, not against this PR: the shared GraphQL tree's 3-level depth bounds VEX detection (acknowledged in the PR description), and gemara's `UpdateAggregateResult` has no `NotApplicable` case, so the `IsActive` step doesn't actually gate — identical to every sibling plan.

data/vex.go
vinayada1: Good catch on the false-positive surface — confirmed `tools/vex/generate.json`, `.vex/data.json`, and `my-vex-tool/config.json` all matched. Fixed in 5e0b1b1, though I went a slightly different route than the suggested one-liner.

Scoping to the immediate parent (`path.Base(dir)`) doesn't actually close the cited cases: in `tools/vex/generate.json`, `.vex/data.json`, and `my-vex-tool/config.json` the immediate parent **is** the vex-token segment, so all three still return true. It also regresses legitimate nested layouts like `security/vex/2024/product.json` → NeedsReview.

Instead I match on **whole path segments** (`hasVexDirSegment`: a segment `== "vex"` or `".vex"`):
- `my-vex-tool/config.json` → false (the real false positive)
- `security/vex/2024/product.json` → true (nested layout preserved)
- `vex/bom.json`, `docs/vex/product.json`, `.vex/data.json` → still true

Note `tools/vex/generate.json` still matches by design — it lives in a directory literally named `vex/`, which is the intended "dedicated VEX directory" convention. Distinguishing a real `vex/` dir from a tooling `vex/` dir isn't possible from the path alone, and since VM-04.02 only ever returns Passed/NeedsReview (never Failed), erring toward Passed there is acceptable. Added test cases for `my-vex-tool/config.json`, `tools/vex-generator/output.json`, and `security/vex/2024/product.json`.

data/payload.go
vinayada1: Agreed — this broke the observed-vs-unknown pattern the PR establishes for `SecurityAdvisories.Known`. Fixed in 5e0b1b1: `HasVexDocument` now checks `payload.Binaries.Err` (already populated from `treeErr` by `analyzeBinaries`) and returns a distinct "Could not scan the repository tree for VEX documents…" message. The verdict stays NeedsReview, but the evidence no longer asserts a scan that didn't fully happen. Added a test covering the tree-error branch.

data/vex.go
vinayada1: Confirmed — `openvexation.json` returned true. Fixed in 5e0b1b1 by folding the marker into the bounded token regex: `(?i)(^|[/._\-])(openvex|vex)([/._\-]|$)`. Now `openvexation.json` → false, while `openvex.json`/`product.openvex.json` stay true. The `cdx-vex`/`cyclonedx-vex` forms are already covered by the delimited `vex` token (the hyphen is a delimiter), so I dropped the separate raw `strings.Contains` branch entirely. Added `openvexation.json` to the test table.

jmeridth (approved):
:robot: (low, non-blocking) Approving at 668a6bf. Verified all three review comments were addressed by execution: whole-segment vex-directory matching excludes tooling dirs (my-vex-tool/, vex-generator/) while preserving nested layouts, and the author's rationale for keeping literal vex/ directories as a pass, plus the pushback on my immediate-parent suggestion, is correct. The bounded (openvex|vex) token closes the openvexation.json gap, and the new VexDocumentsErr field restores the observed-vs-unknown distinction with found documents still passing on partial trees. Rebased onto current main; full suite and -race pass locally.

PR #423, opened 2026-08-04
feat(OSPS-VM-05): evaluate SCA policy, pre-release scanning, and enforcement

### LINKED ISSUE #35: Implement OSPS-VM-05

- [ ] While active, the project documentation MUST include a policy that defines a threshold for remediation of SCA findings related to vulnerabilities and licenses.
- [ ] While active, the project documentation MUST include a policy to address SCA violations prior to any release.
- [ ] While active, all changes to the project's codebase MUST be automatically evaluated against a documented policy for malicious dependencies and known vulnerabilities in dependencies, then blocked in the event of violations, except when declared and suppressed as non-exploitable.

vinayada1:
Closes #35

Implements the three OSPS-VM-05 sub-controls, which were previously wired to `NotImplemented` stubs:

| Control | Requirement |
| --- | --- |
| VM-05.01 | A documented threshold for remediating SCA findings (vulnerabilities and licenses) |
| VM-05.02 | A documented policy for addressing SCA violations prior to release |
| VM-05.03 | Every change automatically evaluated, with violations blocked |

## Approach

No Security Insights field records an SCA remediation threshold or a pre-release SCA policy, so these cannot be read directly. Each control evaluates layered evidence — Security Insights first, then GitHub-observable signals: release and pull-request workflows parsed with `actionlint`, required status checks from rulesets and classic branch protection, and dependency-update tooling config. As a result the checks still produce a meaningful verdict on repositories with no `security-insights.yml`, rather than depending on that file alone.

## Result semantics

The main risk in a compliance scanner is claiming a control is met on evidence that doesn't show it, so results are deliberately conservative:

- **Passed** only where evidence is verifiable end to end. For VM-05.03 this requires the SCA scan to be correlated with a required status check, so that *"a scan runs"* is never reported as *"violations block merges"*.
- **NeedsReview** where a signal exists but its content can't be inspected (e.g. a policy document is linked but not readable), and where evidence couldn't be retrieved at all. A workflow-read failure is never reported as a violation — `Failed` asserts something we observed, and we observed nothing for a file we never read. This matches the convention and rationale already documented in `build_release` and `access_control`.
- **Failed** only on an observed absence, consistent with `docs.HasDependencyManagementPolicy` on the same signal.
- **NotApplicable** for inactive repositories, via the existing `IsActive` step.

## Detection notes

Each of these covers a case that produced a wrong verdict during development, and is now pinned by a regression test:

- The `SCA` acronym is matched on **word boundaries**. As a substring it also matches `scan`, `scanner` and `secret-scanning` — and since `secret-scanning` is a real Security Insights tool type in this codebase, substring matching returned `Passed` at `High` confidence for repositories whose only evidence was a secret scanner plus a required check named `secret-scan`.
- Workflow triggers and steps are read from the parsed **actionlint AST** rather than raw file text, so a tool named in a comment or echoed in a script isn't mistaken for a scan. Both `uses:` actions and `run:` CLI invocations count, since most recognized scanners are CLIs.
- Workflows that actionlint reports **non-fatally** (unknown key, missing `runs-on`) are still inspected; it returns a usable tree alongside those findings, and discarding them understated coverage.
- Release detection requires the `release` event or a push filtered to **version-like tags**, so routine tag pushes such as `dev*` or `nightly` aren't read as releases.

## Behavioral impact

Three controls that previously always returned `NotImplemented` now return real verdicts, so evaluation output for these IDs changes for every scanned repository. No existing control logic was modified.

## Testing

Adds `data.NewPayloadWithWorkflowCache` so workflow-content checks can be exercised without a live GitHub client. Tests cover each control's result paths, plus the detection helpers (`hasVersionTagFilter`, `hasSCAAction`, `looksLikeSCA`, `parseWorkflow`) directly — at the step level a helper regression is indistinguishable from any other no-signal path, which is how an earlier tag-matching regression went unnoticed.

Each regression test was verified to fail against the previous behavior, so the assertions discriminate rather than passing vacuously.

`go build`, `go vet`, `go test ./...`, `golangci-lint` (0 issues), and `make -B build` all pass.

### DIFF
--- a/data/payload.go
+++ b/data/payload.go
@@ -53,7 +53,10 @@ type BinaryAnalysis struct {
 type payloadCache struct {
 	workflows []WorkflowFile
 	// set once workflows have been fetched, so an empty result is not refetched
-	workflowsLoaded bool
+	workflowsLoaded     bool
+	documentation       []DocumentationFile
+	documentationErr    error
+	documentationLoaded bool
 }
 
 // AddEvidence, GetEvidence, and ClearEvidence implement gemara.HasEvidence.
@@ -231,6 +234,7 @@ func (p *Payload) GetWorkflowFiles() ([]WorkflowFile, error) {
 	if p.GraphqlRepoData == nil || p.Config == nil || p.cache == nil {
 		return nil, fmt.Errorf("payload missing required repository data")
 	}
+
 	if p.cache.workflowsLoaded {
 		return p.cache.workflows, nil
 	}
@@ -241,4 +245,24 @@ func (p *Payload) GetWorkflowFiles() ([]WorkflowFile, error) {
 	p.cache.workflows = files
 	p.cache.workflowsLoaded = true
 	return files, nil
+}
+
+// GetDocumentationFiles returns repository documentation once per payload.
+// VM-05 has three controls that inspect the same documents, so caching avoids
+// repeating a potentially large sequence of GitHub content requests. Unlike
+// GetWorkflowFiles, errors are cached deliberately: the walk is expensive and
+// returns whatever partial files it gathered alongside the error, so all three
+// controls observe the same partial evidence rather than re-running the walk.
+func (p *Payload) GetDocumentationFiles() ([]DocumentationFile, error) {
+	if p.RestData == nil || p.cache == nil {
+		return nil, fmt.Errorf("payload missing required repository data")
+	}
+	if p.cache.documentationLoaded {
+		return p.cache.documentation, p.cache.documentationErr
+	}
+	files, err := p.RestData.GetDocumentationFiles()
+	p.cache.documentation = files
+	p.cache.documentationErr = err
+	p.cache.documentationLoaded = true
+	return files, err
 }
--- a/data/rest_data_mock.go
+++ b/data/rest_data_mock.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"github.com/google/go-github/v74/github"
+	"github.com/privateerproj/privateer-sdk/config"
 )
 
 type ClientMock struct {
@@ -42,7 +43,7 @@ func (c *ClientMock) Do(req *http.Request) (*http.Response, error) {
 // cache instead of making an API call. Security Insights is initialized to its
 // empty-but-non-nil shape, matching the state Setup leaves it in.
 func NewRestDataWithContents(contents RepoContent) *RestData {
-	r := &RestData{contents: contents}
+	r := &RestData{contents: contents, contentsObserved: true}
 	r.ensureInsightsInitialized()
 	return r
 }
@@ -66,6 +67,32 @@ func NewPayloadWithHTTPMock(base Payload, body []byte, statusCode int, httpErr e
 	return base
 }
 
+// NewPayloadWithWorkflowCache builds a Payload whose workflow cache is
+// pre-seeded with the given files, so that other packages' tests can exercise
+// workflow-content checks (via GetWorkflowFiles) without a live GitHub client.
+// The cache is marked loaded so GetWorkflowFiles returns files as-is, even when
+// empty, without attempting a fetch.
+func NewPayloadWithWorkflowCache(base Payload, files []WorkflowFile) Payload {
+	if base.RestData == nil {
+		base.RestData = &RestData{}
+	}
+	// GetWorkflowFiles requires these to be non-nil before it will read the
+	// cache, so seed empty defaults when the caller left them unset.
+	if base.GraphqlRepoData == nil {
+		base.GraphqlRepoData = &GraphqlRepoData{}
+	}
+	if base.Config == nil {
+		base.Config = &config.Config{}
+	}
+	if base.cache == nil {
+		base.cache = &payloadCache{}
+	}
+	base.cache.workflows = files
+	base.cache.workflowsLoaded = true
+	base.ensureInsightsInitialized()
+	return base
+}
+
 // NewPayloadWithRepoContents builds a Payload whose RestData is backed by the
 // given root and subdirectory listings, so that other packages' tests can
 // exercise contents-based fallbacks (checkFile, FindFile, FindFileInDirs)
@@ -80,6 +107,10 @@ func NewPayloadWithRepoContents(base Payload, root []*github.RepositoryContent, 
 		sub[dir] = RepoContent{Content: entries}
 	}
 	base.contents = RepoContent{Content: root, SubContent: sub}
+	base.contentsObserved = true
+	if base.cache == nil {
+		base.cache = &payloadCache{}
+	}
 	base.ensureInsightsInitialized()
 	return base
 }
--- a/evaluation_plans/osps/vuln_management/steps.go
+++ b/evaluation_plans/osps/vuln_management/steps.go
@@ -1,6 +1,7 @@
 package vuln_management
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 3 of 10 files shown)


jmeridth:
@vinayada1 conflict to resolve after recent merges.

vinayada1:
> @vinayada1 conflict to resolve after recent merges.

Done

vinayada1:
@jmeridth Rebased onto latest main and resolved the conflict — PR is mergeable and all checks are green.

jmeridth:
@vinayada1 another conflict. 😄

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -695,3 +695,446 @@ func HasPrivateVulnerabilityReporting(payload data.Payload) (result gemara.Resul
...
+		lower := strings.ToLower(ctx)
+		if toolName != "" && strings.Contains(lower, toolName) {
+			return ctx
+		}
+		if looksLikeSCACheck(lower) {
+			return ctx
+		}
+	}
+	return ""
+}
```
jmeridth: :robot: (high) The correlation here is looser than the doc comment claims. `looksLikeSCACheck` matches any context containing the substring `dependency` (via the bare keyword in `scaToolKeywords`), so unrelated required checks are credited as the SCA merge gate. Verified by execution with an SI-declared `grype` tool in CI: required checks named `verify-dependency-lockfile`, `dependency-graph-submission`, and `update-go-dependency-versions` each yield `Passed` at High confidence with the message claiming that check "blocks merges on violations". Those are lockfile-sync and automation checks, not vulnerability gates, so this is the same false-positive class the PR fixed for `sca`/`scan`, relocated to `dependency`. The `toolName` branch has the same issue for short or generic tool names since it is also a `Contains` match.

The sibling `requiredCheckMatchesSast` on main solves this with exact normalized matching against detected sources. Suggest correlating the same way (exact match against the detected workflow job or tool name), or at minimum dropping the bare `dependency` keyword from check-context matching so only concrete tool names and the bounded `sca` acronym qualify.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -695,3 +695,446 @@ func HasPrivateVulnerabilityReporting(payload data.Payload) (result gemara.Resul
...
+	if err != nil {
+		return "", false, err
+	}
+	defaultBranch := ""
+	if payload.GraphqlRepoData != nil {
+		defaultBranch = payload.Repository.DefaultBranchRef.Name
+	}
+	uncertainCandidate := ""
+	for _, file := range workflows {
+		if file.Truncated || (!strings.HasSuffix(file.Name, ".yml") && !strings.HasSuffix(file.Name, ".yaml")) {
```
jmeridth: :robot: (medium) Truncated files (here) and unparseable YAML (the `parseWorkflow` nil check below) are silently skipped, and the same happens in `releaseScaWorkflow` at line 964. Verified by execution: a payload whose only workflow file is truncated returns `Failed` from both `EnforcesSCAOnChanges` ("No SCA tooling ... was found") and `HasSCAReleasePolicy`, asserting an observed absence when nothing was actually inspected. That contradicts the PR's stated rule that a workflow-read failure is never reported as a violation, and it diverges from the `inspectionBlocked` pattern `detectSastInWorkflows` uses for the identical situation in this same file. The new `coverageUncertain` return value is the natural carrier: set it when a workflow file is truncated or fails to parse, and have the callers degrade `Failed` to `NeedsReview` when it is set.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -695,3 +695,446 @@ func HasPrivateVulnerabilityReporting(payload data.Payload) (result gemara.Resul
...
+// vulnerabilities or malicious packages on changes (an SCA gate).
+var dependencyReviewSignatures = []string{
+	"dependency-review-action",
+	"dependency-review",
+	"osv-scanner",
+	"trivy",
+	"grype",
+	"snyk",
+	"dependency-check",
+}
```
jmeridth: :robot: (medium) This list (and `scaCommandIdentifiers` below) misses several mainstream SCA scanners and all license tooling: `npm audit`, `cargo audit`, `pip-audit`, `govulncheck`, `nancy`, `anchore/scan-action` (Grype-based but does not contain the string "grype"), `fossa`, and `licensee` all go undetected (verified by execution). The VM-05.01 catalog text explicitly requires a threshold for SCA findings "related to vulnerabilities and licenses", so a project whose only SCA evidence is FOSSA or licensee config with no Dependabot/Renovate file lands on `Failed` despite having the tooling the control asks about. Under-detection only, no false pass, but a wrong `Failed` is still a wrong definitive verdict. Suggest at least adding the vulnerability CLIs (`pip-audit`, `cargo audit`, `npm audit`, `govulncheck`, `nancy`) and one or two license tools, or degrading the no-signal `Failed` paths to account for the recognized-list being non-exhaustive.

data/rest_data_mock.go
```diff
@@ -66,7 +67,28 @@ func NewPayloadWithHTTPMock(base Payload, body []byte, statusCode int, httpErr e
...
+	}
+	if base.Config == nil {
+		base.Config = &config.Config{}
+	}
+	base.cache = &payloadCache{workflows: files, workflowsLoaded: true}
+	base.ensureInsightsInitialized()
+	return base
+}
+
 // given root and subdirectory listings, so that other packages' tests can
```
jmeridth, line 97: :robot: (low, non-blocking) The insertion of `NewPayloadWithWorkflowCache` deleted the first line of this godoc, so it now starts mid-sentence with no subject:
```suggestion
// NewPayloadWithRepoContents builds a Payload whose RestData is backed by the
// given root and subdirectory listings, so that other packages' tests can
```

jmeridth (changes requested):
:robot: (high) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); every finding was verified by executing the real code, and re-verified against this rebased head before posting.

A lot of this is strong: the word-boundary `sca` acronym matching, the version-tag filter regex, and tolerant `parseWorkflow` all survived extensive adversarial input across three independent review passes, the test suite is genuinely non-vacuous, and the rebase already consolidated onto `workflowCoversChanges` and fixed run-script mention-vs-invocation matching with the shared `normalizeCommand` token approach. Two findings from our earlier internal pass were resolved by those changes before we ever posted them.

Four findings remain (1 high, 2 medium, 1 low non-blocking), detailed in the line comments. The high one matters most: the bare `dependency` keyword lets unrelated required checks (lockfile sync, dependency-graph submission) be credited as the SCA merge gate, producing `Passed` at High confidence, which is the exact property the PR description promises this control avoids.

Housekeeping for the same push: the branch needs a rebase over merged #422; the conflict is mechanical (adjacent insertions in `evaluation-plans.go` and `steps.go`, keep both sides, no symbol overlap or semantic interaction). `SastToolDefined` also became orphaned when main rewired OSPS-VM-06.02, so it can be deleted along the way if that is not already planned.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed: status-check correlation no longer uses the bare `dependency` substring (dropped from the check path) and the tool-name branch now matches at a token boundary, so lockfile-sync/dependency-graph/dependency-update checks are no longer credited as the SCA gate; regression test added and mutation-verified.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed: truncated and unparseable workflows now set the uncertainty signal, and `EnforcesSCAOnChanges`/`HasSCAReleasePolicy` degrade `Failed` to `NeedsReview` when inspection was blocked, matching the `inspectionBlocked` pattern; regression test added and mutation-verified.

evaluation_plans/osps/vuln_management/steps.go
vinayada1: Fixed: added `pip-audit`, `govulncheck`, `nancy`, `cargo audit`, `npm audit` to the CLI list and `anchore/scan-action`, `fossa`, `licensee` to the action/signature list; regression tests added.

data/rest_data_mock.go
vinayada1: Fixed: restored the deleted first line of the `NewPayloadWithRepoContents` godoc.

jmeridth (dismissed):
:robot: (low, non-blocking) Approving at 6ad1e3a. Re-ran every original reproduction against this head: the three dependency-automation check names (verify-dependency-lockfile, dependency-graph-submission, update-go-dependency-versions) now yield NeedsReview instead of a false Passed, while legitimate correlations (grype-scan, dependency-review) still pass and generic tool names no longer over-match at token boundaries. Truncated-only workflow payloads now degrade to NeedsReview on both steps via the uncertainty signal. All five new CLI tools (pip-audit, govulncheck, nancy, cargo audit, npm audit) plus anchore/scan-action detect correctly, clone-URL mentions still do not, and the godoc line is restored. Branch is rebased onto current main and build, vet, full tests, and -race all pass locally.

jmeridth:
@vinayada1 can you please merge main so the checks rekick? leftovers from yesterdays github actions outage.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -759,3 +760,962 @@ func HasVexDocument(payload data.Payload) (result gemara.Result, message string,
...
+	allChangesPattern             = regexp.MustCompile(`\b(?:all|every)\s+(?:code\s+)?changes?\b|\bpull requests?\b|\bbefore merg(?:e|ing)\b`)
+	knownVulnerabilityPattern     = regexp.MustCompile(`\bknown vulnerabilit(?:y|ies)\b`)
+	maliciousDependencyPattern    = regexp.MustCompile(`\bmalicious\b.{0,40}\b(?:dependencies|dependency|packages?|components?)\b|\b(?:dependencies|dependency|packages?|components?)\b.{0,40}\bmalicious\b`)
+	mergeBlockingPattern          = regexp.MustCompile(`\b(?:block|reject|prevent|prohibit)(?:s|ed|ing)?\b.{0,80}\b(?:merg(?:e|ing)|change|pull request|violation)\b|\b(?:merg(?:e|ing)|change|pull request|violation)\b.{0,80}\b(?:block|reject|prevent|prohibit)(?:s|ed|ing)?\b|\bmust pass\b`)
+	nonExploitablePattern         = regexp.MustCompile(`\b(?:non[- ]exploitable|not exploitable|not affected)\b`)
+	exceptionPattern              = regexp.MustCompile(`\b(?:declar(?:e|ed)|suppress(?:ed|ion)?|waiv(?:e|ed|er)|exception|justif(?:y|ied|ication))\b`)
+	scopeNegationPattern          = regexp.MustCompile(`\b(?:not all|not every|only some|only selected)\s+(?:code\s+)?changes?\b`)
+	sentenceSplitPattern          = regexp.MustCompile(`[.!?;]+(?:\s+|$)`)
```
jmeridth, line 783: :robot: (medium) The prose analyzer credits contradictory policy text as `Passed` at High confidence, verified end to end by two independent passes. Three reproductions through `HasSCAReleasePolicy` with a single docs file:

- "SCA findings must be fixed before release; this is currently optional." passes because the `;` here splits the disclaimer into its own statement, detaching it from the obligation clause it modifies.
- "SCA findings must be ignored, not fixed before release." passes because `statementNegated`'s phrase list has no "not <verb>" form.
- "SCA findings must be fixed before release. This requirement is optional." passes because negation is only checked per statement.

All three produce the same result and confidence as a genuine policy. For a compliance scanner that is a wrong definitive verdict from self-contradictory (or gaming-friendly) prose. Cheap mitigations, any subset helps: drop `;` from the splitter so semicolon clauses stay joined to their sentence, extend `statementNegated` to catch "not" adjacent to remediation verbs, and consider checking negation phrases across the surrounding section (or capping prose-derived passes at Medium confidence) so a nearby disclaimer keeps the result at NeedsReview.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -759,3 +760,962 @@ func HasVexDocument(payload data.Payload) (result gemara.Result, message string,
...
+var scaActionIdentifiers = []scaIdentifier{
+	{"actions/dependency-review-action", "dependency-review", true, true},
+	{"google/osv-scanner-action/osv-scanner-action", "osv-scanner", true, false},
+	{"aquasecurity/trivy-action", "trivy", true, false},
+	{"anchore/scan-action", "grype", true, false},
+	{"snyk/actions/node", "snyk", true, false},
+	{"snyk/actions/python", "snyk", true, false},
+	{"snyk/actions/golang", "snyk", true, false},
+	{"dependency-check/dependency-check-action", "dependency-check", true, false},
+}
```
jmeridth, line 810: :robot: (low, non-blocking) Exact-path matching after stripping `@version` misses two real-world forms (verified by execution): `google/osv-scanner-action@v1` returns no match since only the doubled `google/osv-scanner-action/osv-scanner-action` subdir path is listed, and `snyk/actions/docker` is absent while node/python/golang are covered. False negatives only, the exactness correctly rejects clone URLs and forks, so just worth adding the plain osv-scanner repo root and the docker variant.

evaluation_plans/osps/vuln_management/steps.go
```diff
@@ -759,3 +760,962 @@ func HasVexDocument(payload data.Payload) (result gemara.Result, message string,
...
+		return gemara.NeedsReview, "SCA tooling was observed, but all-change coverage, active failure behavior, and exact required-check enforcement were not all verified", gemara.Medium
+	}
+	if payload.RestData != nil {
+		if configPath := payload.DependencyToolingConfig(); configPath != "" {
+			return gemara.NeedsReview, "Dependency tooling was found (" + configPath + "), but it does not prove merge-blocking SCA evaluation of every change", gemara.Low
+		}
+	}
+	if detection.inspectionBlocked {
+		return gemara.NeedsReview, "SCA evidence could not be completely inspected", gemara.Low
+	}
```
jmeridth: :robot: (low, non-blocking) This branch is provably unreachable: reaching this point requires `coverageProven` to be false (every sub-path of the `coverageProven` block above returns), and the earlier `inspectionBlocked && !coverageProven` guard already returned for that combination, so `inspectionBlocked` is always false here. Its message appears in no test, consistent with dead code. Either delete it, or if an intended condition got dropped in the rewrite, restore that. Factoring the decision matrix into a pure function like `evaluateSastEnforcement` in this same file would make this kind of thing directly unit-testable.

data/payload.go
```diff
@@ -242,3 +246,20 @@ func (p *Payload) GetWorkflowFiles() ([]WorkflowFile, error) {
...
+	if p.RestData == nil || p.cache == nil {
+		return nil, fmt.Errorf("payload missing required repository data")
+	}
+	if p.cache.documentationLoaded {
+		return p.cache.documentation, p.cache.documentationErr
+	}
+	files, err := p.RestData.GetDocumentationFiles()
+	p.cache.documentation = files
+	p.cache.documentationErr = err
+	p.cache.documentationLoaded = true
```
jmeridth, line 266: :robot: (low, non-blocking) `documentationLoaded` is set unconditionally, so a transient fetch failure is cached for the rest of the run and reused by all three VM-05 steps, while `GetWorkflowFiles` above only caches on success and retries after an error. Probably intentional given the walk is expensive, and no wrong verdict results (partial files plus the error are preserved), but the two caches in the same struct now have opposite error semantics. Worth a sentence in the comment saying error caching is deliberate, or aligning the two.

jmeridth (changes requested):
:robot: (medium) Multi-model re-review of the rewrite (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); every finding was verified by execution, and the one blocking finding was reproduced independently twice.

This rewrite is a substantial upgrade and most of it verified clean: required-check correlation is now exact and AST-derived (the stale generic-context false-positive class is gone, pinned by tests), app-pinned producers are treated as uncertain rather than trusted by name, truncated and unparseable workflows route to NeedsReview everywhere, no false-Failed path was found in any of the three decision matrices, documentation analysis reads only the authenticated contents API with file and depth caps (no URL fetching), the regex battery is RE2-safe with no pathological inputs found, and all four previously fixed behaviors still hold. The suppression-detection test coverage (`|| true`, `continue-on-error`, `if: false`, needs chains) is genuinely good.

One medium blocks: the policy-prose analyzer produces `Passed` at High confidence for self-contradictory text (semicolon disclaimers, "not <verb>" negations, next-sentence disclaimers), detailed with reproductions and fix options in the line comment on `sentenceSplitPattern`. Three low non-blocking notes accompany it (two missing action identifiers, a provably dead branch, cache error-semantics divergence).

One refutation for the record: a reported case-sensitivity bug in the section classifiers was checked and is not real; sections are lowercased before reaching those helpers.

Follow-up worth an issue rather than this PR: by contrast with the new SCA gating, the merged SAST twin still accepts an SI self-declaration plus a name-coincident required check as Passed with zero workflow evidence, and does not use the new producer-pin protection. Happy to open that issue with the reproduction if useful.

evaluation_plans/osps/vuln_management/steps.go
jmeridth: :robot: (informational) Verified fixed at f15c8e4: the section-level disclaimer scan plus negatedRemediationPattern downgrade all three reproductions to NeedsReview with an accurate contradiction message, while a genuine policy still passes at High. Resolving.

evaluation_plans/osps/vuln_management/steps.go
jmeridth: :robot: (informational) Verified fixed at f15c8e4: google/osv-scanner-action@v1 and snyk/actions/docker both match now. Resolving.

evaluation_plans/osps/vuln_management/steps.go
jmeridth: :robot: (informational) Verified fixed at f15c8e4: the unreachable branch was removed. Resolving.

data/payload.go
jmeridth: :robot: (informational) Verified addressed at f15c8e4: error caching is now documented as deliberate with the partial-evidence rationale. Resolving.

jmeridth (approved):

PR #426, opened 2026-08-06
feat(OSPS-SA-03): implement security assessment and threat modeling checks

### LINKED ISSUE #31: Implement OSPS-SA-03

- [ ] When the project has made a release, the project MUST perform a security assessment to understand the most likely and impactful potential security problems that could occur within the software.
- [ ] When the project has made a release, the project MUST perform a threat modeling and attack surface analysis to understand and protect against attacks on critical code paths, functions, and interactions within the system.

vinayada1:
## Summary

Implements OSPS-SA-03, replacing the `NotImplemented` placeholders for both assessment requirements:

- **OSPS-SA-03.01** (`HasSecurityAssessment`): when the project has made a release, it MUST perform a security assessment to understand the most likely and impactful potential security problems.
- **OSPS-SA-03.02** (`HasThreatModelAnalysis`): when the project has made a release, it MUST perform threat modeling and attack surface analysis.

## Approach

Both checks derive evidence from Security Insights `repository.security.assessments` (self and third-party):

- **Release-gated**: both requirements apply only "when the project has made a release". No published (non-draft) release → `NotApplicable`; unreadable release data → `NeedsReview`.
- **SA-03.02** additionally scans the assessment name/comment for threat-modeling / attack-surface terminology (threat model, attack surface, STRIDE, PASTA, DREAD, attack tree).
- Access to `Insights.Repository` is nil-safe, matching the existing `vuln_management` guard pattern.

## Why `NeedsReview` instead of `Passed` when an assessment is found

Detecting an assessment tells us only that **an artifact exists or is declared** — it does **not** verify that the artifact is correct, current, or sufficient. The scanner cannot, without inspecting the contents, confirm that:

- **SA-03.01**: the assessment actually identifies the *most likely and impactful* security problems for this software.
- **SA-03.02**: the threat model / attack-surface analysis genuinely covers the *critical code paths, functions, and interactions*, and enumerates threats with mitigations.

A keyword match (e.g. finding "STRIDE" or "threat model") or a populated Security Insights field is trivially satisfiable and could be a stub, an outdated document, or a checkbox entry. Auto-passing on that basis would let projects clear a MUST requirement without any substantive analysis, giving downstream consumers false assurance. Conversely, failing a project that *has* done the work but expressed it in unrecognized terms would be wrong.

`NeedsReview` is the correct verdict because sufficiency is a **judgment about document contents** that today requires a human (or a future AI-assisted evaluator). We therefore return:

- `NeedsReview` when an assessment is found → a reviewer confirms coverage before it counts as a pass.
- `Failed` when the project has released but no assessment evidence exists at all.

This mirrors how other non-deterministic content checks already behave in this repo (e.g. `TestExecutionDocumentation` / QA-06.02 and `DocumentsTestMaintenancePolicy` / QA-06.03 defer to manual review rather than auto-passing), and it deliberately leaves a clean seam to plug in AI-assisted content evaluation later, at which point qualifying artifacts could be promoted to `Passed`.

## Result matrix

| Condition | 03.01 | 03.02 |
|---|---|---|
| No published release | NotApplicable | NotApplicable |
| Release data unreadable | NeedsReview | NeedsReview |
| Assessment declared / found | NeedsReview | NeedsReview |
| Released, no assessment | Failed | Failed |

## Testing

- Table-driven tests for both steps with exact result + message assertions across every branch (release gating, draft-only releases, self/third-party assessments, empty/whitespace-only assessments, case-insensitive keyword matching, nil `Insights.Repository`).
- Package coverage: 100% of statements.
- `go test ./...`, `go vet`, and `golangci-lint` all pass.

Closes #31

### DIFF
--- a/evaluation_plans/osps/sec_assessment/steps.go
+++ b/evaluation_plans/osps/sec_assessment/steps.go
@@ -1,11 +1,13 @@
 package sec_assessment
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/gemaraproj/go-gemara"
 	"github.com/ossf/pvtr-github-repo-scanner/data"
 	"github.com/ossf/pvtr-github-repo-scanner/evaluation_plans/reusable_steps"
+	"github.com/ossf/si-tooling/v2/si"
 )
 
 // DesignDocFiles are common file names for design/architecture documentation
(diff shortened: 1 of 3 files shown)


jmeridth:
@vinayada1 can you please merge main? need to rekick checks after github actions outage yesterday.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +67,148 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+			return true, true
+		}
+	}
+	return false, true
+}
+
+// assessmentIsPopulated reports whether a Security Insights assessment carries
+// any evidence of having been performed. Comment is a required field in the
+// schema, so an all-zero Assessment means none was declared.
+func assessmentIsPopulated(assessment si.Assessment) bool {
```
jmeridth: :robot: (medium) `Comment` is a required SI field that real-world files populate even to declare the opposite of an assessment. si-tooling's own `security-insights.yml` says `self: comment: No self assessment completed`, and this repository's own says `A formal self-assessment has not yet been completed for this project.` Both count as populated here, so both steps report NeedsReview with "Security Insights declares a self security assessment", while the PR's result matrix defines that state as Failed. Verified by executing the step against this repo's real SI content: the scanner misclassifies its own repository, with a message asserting the reverse of what the file says.

Suggested direction: before crediting a comment-only assessment, screen the comment for negative phrasing ("not", "no ", "has not been", "never"), or require `name`/`evidence` alongside a bare comment; the negation patterns added in the VM-05 work are precedent. At minimum the message should not claim a declaration that the text explicitly denies.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +67,148 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+func HasSecurityAssessment(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	released, observable := hasPublishedRelease(payload)
+	if !observable {
+		return gemara.NeedsReview, "Release data is unavailable; manually review whether a security assessment was performed", gemara.Low
+	}
+	if !released {
+		return gemara.NotApplicable, "No published releases found; the security-assessment requirement does not apply", gemara.High
+	}
+
+	assessments := securityAssessments(payload).Assessments
```
jmeridth, line 291: :robot: (medium) Neither step consults `payload.InsightsError`, so a Security Insights file that exists but fails to parse produces a confident `Failed` (verified by execution: `RestData{InsightsError: true}` plus a published release returns Failed at Medium for both steps). The strict YAML load amplifies this: one stray key anywhere makes the whole file unparseable, so a project that declares its assessments correctly can Fail over an unrelated typo. House convention treats this as inconclusive (`reusable_steps.HasSecurityInsightsFile` returns NeedsReview on `InsightsError`, and `SastEnforcedOnChanges` marks inspection blocked). An `if payload.InsightsError { return gemara.NeedsReview, ..., gemara.Low }` after the release gate in both steps restores the pattern.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +67,148 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+	}
+	if assessment.Evidence != nil && strings.TrimSpace(string(*assessment.Evidence)) != "" {
+		return true
+	}
+	return false
+}
+
+// mentionsThreatModeling reports whether an assessment's name or comment
+// references threat modeling or attack surface analysis.
+func mentionsThreatModeling(assessment si.Assessment) bool {
```
jmeridth, line 246: :robot: (low, non-blocking) Two related notes on this scan, both message-only since the matched and unmatched branches each return NeedsReview at Low, so the keywords can never change a verdict. First, substring matching means "restrided", "antipasta", and "dreadful" all count as threat-modeling mentions. Second, with no negation handling, a comment reading "no threat modeling was performed" produces the message "Security Insights declares threat modeling or attack surface analysis", the reverse of what the file says. If the distinction is meant to carry weight, differentiating the two branches' confidence would make the scan meaningful; otherwise a comment documenting it as message-only phrasing (and softening the "declares" wording for negated text) would keep the emitted evidence accurate. Also noting `Evidence` counts toward populated but is not scanned here, so an evidence URL literally named threat-model.md still gets the "does not mention" message.

jmeridth (changes requested):
:robot: (medium) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); all findings verified by execution, the headline one additionally against this repository's own real security-insights.yml.

Much of this verified clean: the full 17-row result matrix executes exactly as the PR body describes, si-tooling v2.2.0 field usage is correct (round-tripped through real YAML), nil-safety holds on every reachable path, draft and ReleasesError gating matches ReleasesHaveSBOM, catalog text aligns across all three YAMLs, the deliberate no-Passed design has house precedent, and the 100% coverage claim is real with exact-message assertions. The rationale in the PR body for NeedsReview over Passed is well argued.

Two mediums block, detailed in line comments: comment-only "no assessment completed" declarations (present in this repo's own SI file and si-tooling's) are credited as declared assessments with a backwards message, and InsightsError is ignored so an unparseable SI file yields a confident Failed.

Non-blocking notes: the SI-only evidence base for Failed matches many existing steps but diverges from sibling SA-01 in this same file, which also scans the repo tree; a cheap root-filename fallback (threat-model.md, security-assessment.md) would reduce false Fails for non-SI projects and is worth considering. Date-only assessments (schema-nonconformant but parseable) count as absent. `securityAssessments`'s doc comment advertises nil tolerance it lacks for nil RestData (unreachable today). README counts were not updated and will collide with #427's update. The `hasPublishedRelease` duplication with #427 is flagged on that PR; coordinating the shared helper before either merges avoids a redeclaration conflict.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Agreed — fixed in d5eeb75. `assessmentIsPopulated` is replaced by `assessmentDeclared`, which credits a structured `Name`/`Evidence` unconditionally but screens a bare `Comment` for negation phrases (`assessmentDenials`: "not been completed", "no self assessment", "has not", "not yet", etc.) before treating it as a declaration. Verified against this repo's own `.github/security-insights.yml` ("A formal self-assessment has not yet been completed...") — SA-03.01 now returns Failed instead of a NeedsReview that asserted the reverse of the file. Added test cases for both real-world phrasings plus the "negated comment but populated Name still declares" case.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Agreed — fixed in d5eeb75. Both steps now check `payload.InsightsError` immediately after the release gate and return `NeedsReview` at `Low` ("Security Insights file could not be parsed; manually review..."), matching the `HasSecurityInsightsFile` precedent. This prevents a confident `Failed` when an unrelated typo makes the strict YAML load fail. Added an `InsightsError: true` test case to both step tests.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Partially addressed in d5eeb75:

- **Negation** ("no threat modeling was performed" → "declares..."): resolved for free by the `assessmentDeclared` fix — a negated bare comment is no longer credited, so it never reaches `mentionsThreatModeling`. Added a negated-comment test case here too.
- **Evidence not scanned**: fixed — `mentionsThreatModeling` now also lowercases and scans `Evidence`, so an evidence URL named `threat-model.md` is recognized. Added a test for that.
- **Substring false-positives** ("dreadful", "antipasta", "restrided"): intentionally **not** changed. Both the matched and unmatched branches return `NeedsReview`/`Low`, so these can only affect the message string, never the verdict. Tightening to word boundaries risks dropping legitimate methodology names (STRIDE/PASTA/DREAD appear inside compound words in practice) for zero verdict benefit. Tracking as a separate low-priority polish issue rather than expanding this PR's surface.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: (medium) Verified the fix against both real-world files and confirmed they now correctly produce Failed, and the InsightsError and Evidence changes check out too. One regression though, reproduced by execution at d5eeb75: the denial list's broad fragments ("has not", "have not", "not yet") match anywhere in a comment, so a genuine comment-only declaration like "Self assessment completed in 2024; scope has not changed since." now returns Failed for both steps. A true declaration flipping to a definitive wrong verdict is a worse direction than the original bug.

Trimming the broad fragments fixes it: the specific phrases ("not been completed", "not performed", "not conducted", "no self assessment", "no assessment", "no formal") already cover this repo's own SI file and si-tooling's. While in there: "no threat modeling was performed" (the phrasing from the original comment) still is not caught by any denial phrase, so it is credited as a declaration with the backwards message; a "no <activity> was performed" style pattern or adding "was not performed" and "no threat model" would close that too.

jmeridth:
@vinayada1 comment on a medium issue and conflicts to resolve.

jmeridth:
:robot: (informational) Thanks for the quick rebase over #427 and for consolidating onto the shared `reusable_steps.HasPublishedRelease`, that resolved the conflict cleanly and is exactly the structure we hoped for. One item is still open though: the follow-up on the `assessmentDeclared` thread (the broad denial fragments like "has not" flipping genuine comment-only declarations to Failed) looks like it crossed paths with your push, so it may not have been on your radar yet. The request-changes review stands on that one point; happy to re-verify and approve as soon as it lands.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Good catch — agreed, and fixed. The broad fragments (`"has not"`, `"have not"`, `"hasn't"`, `"haven't"`, `"not yet"`) are removed from `assessmentDenials`; only specific phrases remain (`"not been completed"`, `"not yet been completed"`, `"not performed"`, `"not conducted"`, `"no self assessment"`, `"no assessment"`, `"no formal"`, `"never been/performed/conducted"`, etc.).

Verified by execution:
- `"Self assessment completed in 2024; scope has not changed since."` → now **NeedsReview** (credited as a declaration), no longer Failed.
- si-tooling's `"No self assessment completed"` and this repo's `"...has not yet been completed"` → still correctly **Failed**.

Added a regression test (`"genuine declaration with incidental 'has not' is credited"`) asserting the NeedsReview result + message. 100% coverage retained.

On the second point — `"no threat modeling was performed"` producing the reversed "declares threat modeling" message: I've intentionally left the shared `assessmentDenials` list free of threat-model-specific phrases. It gates both steps on whether a *security assessment* exists, so adding `"no threat model"` there would wrongly Fail **SA-03.01** for a legitimate comment like "Full security assessment done; no threat model created separately." Since the threat-scan branches are both NeedsReview/Low, this is message-only and never changes a verdict. Happy to handle it narrowly inside the threat scan if you'd prefer, but kept it out of the verdict-bearing shared path here.

jmeridth: :robot: (informational) Verified fixed at a846151 by execution: the mixed-statement reproduction now credits as a declaration (NeedsReview), both real-world denial comments still correctly produce Failed, and "was not performed" phrasing is caught. The lone residue ("no threat modeling was performed" matching no denial phrase) is lenient-direction and non-blocking. Resolving.

jmeridth (approved):
:robot: (low, non-blocking) Approving at a846151. All three original findings plus the follow-up regression are verified fixed by execution: negation-aware assessment declarations handle both real-world denial comments (this repo's own SI file included) while genuine and mixed-statement declarations credit correctly, InsightsError routes to NeedsReview in both steps, and Evidence URLs are scanned for threat-modeling terms. The rebase onto the shared reusable_steps.HasPublishedRelease resolved the #427 duplication cleanly. Full suite passes locally. One accepted lenient-direction residue is noted in the resolved thread.

PR #427, opened 2026-08-06
feat(OSPS-SA-02): implement external interface documentation check

### LINKED ISSUE #30: Implement OSPS-SA-02

- [ ] When the project has made a release, the project documentation MUST include descriptions of all external software interfaces of the released software assets.

vinayada1:
## Summary

Implements **OSPS-SA-02.01** (closes #30): when the project has made a release, the project documentation MUST include descriptions of all external software interfaces of the released software assets.

Previously `OSPS-SA-02.01` was wired to `reusable_steps.NotImplemented`.

## Approach

Follows the OSPS-SA-03 convention from #426: because automation cannot confirm that documentation covers *all* external interfaces, any positive signal returns **NeedsReview** rather than **Passed**.

New step `sec_assessment.HasExternalInterfaceDocumentation`:
- **NeedsReview** when release data is unobservable (nil `RestData` / releases error)
- **NotApplicable** when there is no published (non-draft) release — the requirement is release-gated
- **NeedsReview** when a positive signal is found:
  - a root API/spec file — exact, case-insensitive names (no globbing): `api.md/rst/txt/yaml/yml/json`, `apidocs.md`, `api-reference.md/rst`, `openapi.yaml/yml/json`, `swagger.yaml/yml/json`
  - a directory that commonly holds API docs (`api`, `apis`, `apidocs`, `api-docs`, `documentation`, `reference`, `references`, `docs`, `doc`, `spec`, `specs`, `openapi`, `swagger`, `schema`, `proto`)
  - a `DetailedGuide` / `QuickstartGuide` declared in Security Insights
- **Failed** (`gemara.Medium`) when a release exists but no such evidence is found

Release gating uses the shared `reusable_steps.HasPublishedRelease` helper (extracted so it is reused rather than duplicated against #426).

## Review feedback addressed

- **Extracted `hasPublishedRelease` → `reusable_steps.HasPublishedRelease`** to remove the byte-identical duplication with the still-open #426 that would otherwise cause a `redeclared in this block` compile error for whichever PR merged second. #426 can consolidate onto the shared helper.
- **Fixed the leaked `Undetermined` confidence** on the terminal `Failed` branch — it now returns an explicit `gemara.Medium` (asserted in tests).
- **Widened the file/dir lists** (`api.yaml/yml/json`; `documentation`, `spec`, `specs`, `openapi`, `swagger`, `schema`, `proto` directories) to avoid false `Failed` results on common legitimate layouts, while keeping a reachable `Failed` consistent with the repo's DO-* / SA-01 "missing required docs → Failed" convention.
- **Clarified messages** to say "file or directory" where both scans came up empty.

## Testing

- `make build`, `go vet`, and `go test ./evaluation_plans/...` all pass
- **100%** statement coverage on `HasExternalInterfaceDocumentation` and `reusable_steps.HasPublishedRelease`

## Docs

- README control counts updated (41 → 42 covered, 11 → 10 not implemented)

No catalog YAML changes needed — `OSPS-SA-02.01` is already defined in the active catalogs.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/evaluation_plans/osps/sec_assessment/steps.go
+++ b/evaluation_plans/osps/sec_assessment/steps.go
@@ -65,3 +65,105 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
 
 	return gemara.Failed, "Design documentation demonstrating all actions and actors was NOT found", confidence
 }
+
+// InterfaceDocFiles are common file names for external interface / API documentation.
+var InterfaceDocFiles = []string{
+	"api.md",
+	"api.rst",
+	"api.txt",
+	"apidocs.md",
+	"api-reference.md",
+	"api-reference.rst",
+	"openapi.yaml",
+	"openapi.yml",
+	"openapi.json",
+	"swagger.yaml",
+	"swagger.yml",
+	"swagger.json",
+}
+
+// InterfaceDocDirectories are common directory names that typically contain
+// external interface / API documentation.
+var InterfaceDocDirectories = []string{
+	"api",
+	"apis",
+	"apidocs",
+	"api-docs",
+	"reference",
+	"references",
+	"docs",
+	"doc",
+}
+
+// hasPublishedRelease reports whether the project has published at least one
+// non-draft release. observable is false when release data could not be seen,
+// so callers can degrade to manual review instead of a definitive result.
+func hasPublishedRelease(payload data.Payload) (released bool, observable bool) {
+	// Missing REST data or a fetch error means release state is unknown.
+	if payload.RestData == nil || payload.ReleasesError != nil {
+		return false, false
+	}
+	for _, release := range payload.Releases {
+		if !release.Draft {
+			return true, true
+		}
+	}
+	return false, true
+}
+
+// HasExternalInterfaceDocumentation assesses OSPS-SA-02.01: when the project has
+// made a release, its documentation must describe all external software
+// interfaces (APIs) of the released assets.
+func HasExternalInterfaceDocumentation(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	// The requirement only applies once a release exists.
+	released, observable := hasPublishedRelease(payload)
+	if !observable {
+		return gemara.NeedsReview, "Release data is unavailable; manually review whether the documentation describes all external software interfaces", gemara.Low
+	}
+	if !released {
+		return gemara.NotApplicable, "No published releases found; the external interface documentation requirement does not apply", gemara.High
+	}
+
+	var foundDirectories []string
+
+	if payload.GraphqlRepoData != nil {
+		for _, entry := range payload.Repository.Object.Tree.Entries {
+			if entry.Type == "blob" {
+				for _, docFile := range InterfaceDocFiles {
+					if strings.EqualFold(entry.Name, docFile) {
+						// A matching file proves interface docs exist, but not that
+						// they cover every external interface of the released assets.
+						return gemara.NeedsReview, "External interface documentation found (" + entry.Name + "), but coverage of all external interfaces requires manual review", gemara.Low
+					}
+				}
+			}
+
+			if entry.Type == "tree" {
+				for _, docDir := range InterfaceDocDirectories {
+					if strings.EqualFold(entry.Name, docDir) {
+						foundDirectories = append(foundDirectories, entry.Name)
+					}
+				}
+			}
+		}
+	}
+
+	// A directory that typically holds API docs is a weaker signal that still
+	// cannot be confirmed to document every interface.
+	if len(foundDirectories) > 0 {
+		return gemara.NeedsReview, "No external interface documentation file found in root, but found directories that may contain API documentation: " + strings.Join(foundDirectories, ", ") + " - manual review needed to confirm all external interfaces are documented", gemara.Low
+	}
+
+	// Fallback: a detailed or quickstart guide in Security Insights may describe
+	// the interfaces, but this cannot be verified automatically.
+	if payload.RestData != nil && payload.Insights.Project != nil && payload.Insights.Project.Documentation != nil {
+		if payload.Insights.Project.Documentation.DetailedGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but detailed guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
+		}
+		if payload.Insights.Project.Documentation.QuickstartGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but quickstart guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
+		}
+	}
+
+	return gemara.Failed, "Documentation describing the external software interfaces of released assets was NOT found", confidence
+}
--- a/evaluation_plans/osps/sec_assessment/steps.go
+++ b/evaluation_plans/osps/sec_assessment/steps.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gemaraproj/go-gemara"
 	"github.com/ossf/pvtr-github-repo-scanner/data"
+	"github.com/ossf/pvtr-github-repo-scanner/evaluation_plans/reusable_steps"
 )
 
 // DesignDocFiles are common file names for design/architecture documentation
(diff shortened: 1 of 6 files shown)


jmeridth:
@vinayada1 can you please merge main? need to rekick checks after github actions outage yesterday.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +65,105 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+	"reference",
+	"references",
+	"docs",
+	"doc",
+}
+
+// hasPublishedRelease reports whether the project has published at least one
+// non-draft release. observable is false when release data could not be seen,
+// so callers can degrade to manual review instead of a definitive result.
+func hasPublishedRelease(payload data.Payload) (released bool, observable bool) {
```
jmeridth: :robot: (medium) Coordination needed with #426, which is still open (the PR body describes it as merged): both PRs add a byte-identical `hasPublishedRelease` with the same signature to this same file, and both append tests at the same tail of `steps_test.go`. Two reviewers reproduced the merge independently: content conflicts in both files, and a keep-both resolution fails to compile with `hasPublishedRelease redeclared in this block`. Whichever PR lands second breaks unless carefully resolved. Suggest either extracting the helper somewhere shared (`reusable_steps`?) before either merges, or agreeing on merge order and having the second PR rebase to reuse the landed copy.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +65,105 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+	if payload.RestData != nil && payload.Insights.Project != nil && payload.Insights.Project.Documentation != nil {
+		if payload.Insights.Project.Documentation.DetailedGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but detailed guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
+		}
+		if payload.Insights.Project.Documentation.QuickstartGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but quickstart guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
+		}
+	}
+
+	return gemara.Failed, "Documentation describing the external software interfaces of released assets was NOT found", confidence
```
jmeridth: :robot: (medium) This one definitive verdict rests on the weakest evidence in the function, and it also ships confidence `Undetermined` (the named return is never assigned; the tests discard confidence so nothing catches it). Verified by probe: README-only API docs, a `documentation/`, `spec/`, `proto/`, or `openapi/` directory, `api.yaml`, `openapi-spec.yaml`, and `api.apib` all land here and Fail, while an empty `docs/` gets NeedsReview. Every other branch follows the stated principle that automation cannot confirm coverage; applying the same principle to the no-evidence case avoids false Fails for the very common README-documented library and fixes the confidence gap in the same line:
```suggestion
	return gemara.NeedsReview, "No external interface documentation file, API-doc directory, or Security Insights guide was found; manually review whether the documentation describes all external software interfaces of released assets", gemara.Low
```
If you would rather keep a reachable Failed, widening the lists (at least `documentation`, `spec`, `specs`, `openapi`, `swagger`, `schema`, `proto` directories and `api.yaml`/`api.json`) plus an explicit `gemara.Medium` confidence would be the alternative.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -65,3 +65,105 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
+	}
+
+	// Fallback: a detailed or quickstart guide in Security Insights may describe
+	// the interfaces, but this cannot be verified automatically.
+	if payload.RestData != nil && payload.Insights.Project != nil && payload.Insights.Project.Documentation != nil {
+		if payload.Insights.Project.Documentation.DetailedGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but detailed guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
+		}
+		if payload.Insights.Project.Documentation.QuickstartGuide != nil {
+			return gemara.NeedsReview, "No external interface documentation file found, but quickstart guide specified in Security Insights - manual review needed to confirm all external interfaces are documented", gemara.Low
```
jmeridth: :robot: (low, non-blocking) Small message-accuracy nit: by the time these branches run, both the file scan and the directory scan have come up empty, but the messages only say "No external interface documentation file found". Saying "file or directory" keeps the emitted evidence accurate. Same spirit for the comment at the file-match branch above: a matching root filename indicates interface docs likely exist rather than proving it.

jmeridth (changes requested):
:robot: (medium) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5 agents plus a Fable 5 synthesis pass); every finding verified by execution or a real merge simulation.

Most of this checks out cleanly: the matchers are exact and case-insensitive with correct blob/tree discrimination and precedence (55 adversarial cases, zero failures), nil-safety is better than the adjacent SA-01.01 step, draft and ReleasesError gating matches the ReleasesHaveSBOM convention, the 100% coverage claim is real, and tests assert exact results and messages.

Two mediums block, detailed in the line comments: the `hasPublishedRelease` duplication against still-open #426 (whichever merges second hits a conflict that resolves into a redeclaration compile error), and the terminal `Failed` branch, which fires on common legitimate layouts (README-documented APIs, `documentation/`, `spec/`, `proto/`, `openapi/` dirs, `api.yaml`) while also shipping an unset `Undetermined` confidence.

Noted for awareness, not blocking: partial GraphQL data that zeroes the tree falls through to Failed (shared convention across seven steps, hard to hit in practice); tag-only releases yield NotApplicable at High confidence (project-wide convention worth a global rethink someday); SA-01.01 and SA-02.01 now disagree on release gating and pass strictness for identical repo states, with this PR on the more defensible side, so aligning SA-01.01 would make a good follow-up; the PR body's `openapi.*` phrasing implies globbing that the code (correctly) does not do.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Agreed — this was a real redeclaration hazard. Fixed in fec3081 by extracting the helper into the shared `reusable_steps` package as exported `HasPublishedRelease`, and removing the local copy here. Now there is no cross-PR redeclaration regardless of merge order: the helper lives in a different package under a different name, so #426 (still open) can either keep its local copy or, preferably, consolidate onto `reusable_steps.HasPublishedRelease` when it rebases. 100% covered by a new `Test_HasPublishedRelease`.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in fec3081, taking your "keep a reachable Failed" alternative:

- **Confidence bug:** the terminal branch no longer leaks the unset named return; it now returns an explicit `gemara.Medium`, asserted directly in tests (added a `checkConf` column so confidence is no longer discarded).
- **False Fails:** widened the lists exactly as you suggested — files now include `api.yaml/yml/json`, and directories add `documentation`, `spec`, `specs`, `openapi`, `swagger`, `schema`, `proto`. So `documentation/`, `spec/`, `proto/`, `openapi/`, `api.yaml`, etc. now correctly land in NeedsReview.

I kept a reachable `Failed` (rather than converting it to NeedsReview) deliberately: it stays consistent with the repository convention for a required-docs MUST — SA-01.01 and the DO-* family (`HasSupportDocs`, `HasUserGuides`, `HasBuildInstructions`, ...) all terminate in `Failed` when no evidence exists. A step that can never Fail would provide little signal for a MUST control. README-only API docs are the remaining gap; parsing README sections (like DO-07 does) is a reasonable future enhancement but felt out of scope here given the false-positive risk.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Agreed, fixed in fec3081. The two Security Insights fallback messages now say "No external interface documentation file or directory found", and the file-match comment was softened to "a matching root filename indicates interface docs likely exist, but does not prove they cover every external interface".

jmeridth (approved):
:robot: (low, non-blocking) Approving at 6160b75. Verified all three review comments by execution: the release-gate helper now lives in reusable_steps.HasPublishedRelease with its own tests and the local copy is removed, dissolving the #426 redeclaration hazard regardless of merge order; the widened lists route documentation/, spec/, proto/, openapi/ directories and api.yaml to NeedsReview while README-only repos keep a reachable Failed at explicit Medium confidence (now asserted in tests); and the fallback messages and comment wording are accurate. Full suite and -race pass locally.

PR #441, opened 2026-08-12
feat: implement OSPS-BR-01.04

### LINKED ISSUE #336: Add OSPS-BR-01.04 support
OSPS-BR-01.04 is present in OSPS Baseline 2026-02-19 but is not fully supported by the scanner yet.

Work needed:

- Add the requirement to the catalog data if it is missing
- Replace the current NotImplemented stub with a real check
- Add focused tests for the new check

Related context: older control-level issue ossf/pvtr-github-repo-scanner#16 was closed before this sub-requirement existed.

Satarupa22-SD:
fixes #336

### DIFF
--- a/evaluation_plans/osps/build_release/steps.go
+++ b/evaluation_plans/osps/build_release/steps.go
@@ -898,3 +898,165 @@ func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Res
 	}
 	return gemara.NeedsReview, "No dependency manifests found in the GitHub dependency graph. Review the project to confirm that any dependencies ingested by the build and release pipeline use standardized tooling.", confidence
 }
+
+// CicdSanitizesCollaboratorInput implements OSPS-BR-01.04: CI/CD pipelines
+// which accept trusted collaborator input MUST sanitize and validate that
+// input prior to use in the pipeline.
+//
+// "Trusted collaborator input" here is workflow_dispatch inputs: values a
+// collaborator with write access types in when manually triggering a
+// workflow. Unlike OSPS-BR-01.01's fixed list of GitHub-controlled untrusted
+// metadata, workflow_dispatch input names are workflow-specific, so this
+// check derives the sensitive expressions from each workflow's own declared
+// inputs, then flags any of them interpolated directly into a run: step's
+// script body. An input assigned to a step-level env: var and referenced as
+// a shell variable is not flagged, matching the same convention used by
+// CicdSanitizedInputParameters (OSPS-BR-01.01): that indirection is the
+// recommended mitigation, not the vulnerability.
+//
+//   - Failed: a declared workflow_dispatch input is interpolated directly
+//     into a run: step's script body.
+//   - NeedsReview: one or more workflow files could not be read (truncated
+//     by the API or unparseable), so an unsanitized input could be hiding
+//     there.
+//   - Passed: every declared workflow_dispatch input is either unused in
+//     run: steps or only referenced via a safer indirection.
+//   - NotApplicable: no workflow declares workflow_dispatch inputs, so there
+//     is no trusted collaborator input for a CI/CD pipeline to sanitize.
+func CicdSanitizesCollaboratorInput(payload data.Payload) (gemara.Result, string, gemara.ConfidenceLevel) {
+	var confidence gemara.ConfidenceLevel
+
+	workflows, err := payload.GetWorkflowFiles()
+	if len(workflows) == 0 {
+		message := "No workflows found in .github/workflows directory"
+		if err != nil {
+			message = err.Error()
+		}
+		return gemara.NotApplicable, message, confidence
+	}
+
+	result, message := evaluateCollaboratorInputSanitization(workflows)
+	return result, message, confidence
+}
+
+// evaluateCollaboratorInputSanitization decodes each workflow file, collects
+// its declared workflow_dispatch input names, and checks run: steps for
+// direct interpolation of those inputs. Separated from payload retrieval so
+// the classification logic is unit-testable without a payload fixture,
+// mirroring evaluateUntrustedCodeIsolation.
+func evaluateCollaboratorInputSanitization(workflows []data.WorkflowFile) (gemara.Result, string) {
+	var violations []string
+	var uninspected []string
+	hasDispatchInputs := false
+
+	for _, file := range workflows {
+		if !strings.HasSuffix(file.Name, ".yml") && !strings.HasSuffix(file.Name, ".yaml") {
+			continue
+		}
+		if file.Truncated {
+			uninspected = append(uninspected, fmt.Sprintf("%s (too large to retrieve)", file.Path))
+			continue
+		}
+		workflow, parseErr := actionlint.Parse([]byte(file.Content))
+		if parseErr != nil {
+			uninspected = append(uninspected, fmt.Sprintf("%s (%v)", file.Path, parseErr))
+			continue
+		}
+
+		inputNames := workflowDispatchInputNames(workflow)
+		if len(inputNames) == 0 {
+			continue
+		}
+		hasDispatchInputs = true
+
+		pattern := unsanitizedDispatchInputPattern(inputNames)
+		violations = append(violations, checkWorkflowFileForUnsanitizedDispatchInputs(file.Name, workflow, pattern)...)
+	}
+
+	if len(violations) > 0 {
+		return gemara.Failed,
+			"CI/CD pipelines use workflow_dispatch inputs without sanitization: " + strings.Join(violations, "; ")
+	}
+
+	if len(uninspected) > 0 {
+		return gemara.NeedsReview, fmt.Sprintf(
+			"Unable to evaluate %d of %d workflow files, manual review required: %s",
+			len(uninspected), len(workflows), strings.Join(uninspected, "; "))
+	}
+
+	if !hasDispatchInputs {
+		return gemara.NotApplicable,
+			"No workflow declares workflow_dispatch inputs; there is no trusted collaborator input for a CI/CD pipeline to sanitize"
+	}
+
+	return gemara.Passed,
+		"All workflow_dispatch inputs are either unused in run: steps or safely indirected (e.g. via env:) rather than interpolated directly"
+}
+
+// workflowDispatchInputNames returns the workflow_dispatch input names
+// declared by a workflow. Names are already lower-cased by actionlint since
+// GitHub Actions input names are case-insensitive.
+func workflowDispatchInputNames(workflow *actionlint.Workflow) []string {
+	var names []string
+	for _, event := range workflow.On {
+		dispatch, ok := event.(*actionlint.WorkflowDispatchEvent)
+		if !ok {
+			continue
+		}
+		for name := range dispatch.Inputs {
+			names = append(names, name)
+		}
+	}
+	return names
+}
+
+// unsanitizedDispatchInputPattern builds a regex matching either the
+// `inputs.<name>` or `github.event.inputs.<name>` context expression for any
+// of the given input names, anchored so e.g. "inputs.tag" does not also match
+// "inputs.tag2".
+func unsanitizedDispatchInputPattern(inputNames []string) *regexp.Regexp {
+	quoted := make([]string, len(inputNames))
+	for i, name := range inputNames {
+		quoted[i] = regexp.QuoteMeta(name)
+	}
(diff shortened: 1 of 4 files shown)


evaluation_plans/osps/build_release/steps.go
```diff
@@ -898,3 +898,165 @@ func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Res
...
+// "inputs.tag2".
+func unsanitizedDispatchInputPattern(inputNames []string) *regexp.Regexp {
+	quoted := make([]string, len(inputNames))
+	for i, name := range inputNames {
+		quoted[i] = regexp.QuoteMeta(name)
+	}
+	return regexp.MustCompile(`(?i)^(github\.event\.)?inputs\.(` + strings.Join(quoted, "|") + `)$`)
+}
+
+// checkWorkflowFileForUnsanitizedDispatchInputs reports every run: step that
```
jmeridth: :robot: (high) The prefix-collision anchoring here is exactly the right instinct (and the `inputs.target` vs `inputs.target2` test proves it works), but the anchors are applied to the whole `${{ }}` expression rather than the token. `pullVariablesFromScript` returns the entire expression body, so any compound expression escapes detection. We verified these all return Passed through the real step, while each one splices collaborator input into the script identically to the bare form:

- `./deploy.sh ${{ inputs.target || 'default' }}` (probably the most common real-world dispatch idiom)
- `./deploy.sh ${{ format('--target={0}', inputs.target) }}`
- `./deploy.sh ${{ inputs.target != '' && inputs.target || 'x' }}`
- `echo '${{ toJSON(github.event.inputs) }}' | ./deploy.sh`

The fix keeps your prefix safety by swapping the anchors for boundary guards, something like:

```go
return regexp.MustCompile(`(?i)(^|[^\w.-])(github\.event\.)?inputs\.(` + strings.Join(quoted, "|") + `)($|[^-\w])`)
```

matched against the expression as today. The `($|[^-\w])` tail is what preserves the `target` vs `target2` distinction, including hyphenated names like `release-tag`. Adding one test row per bullet above would pin it. For comparison, the sibling `untrustedVars` pattern at the top of this file is deliberately unanchored for this same reason.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -898,3 +898,165 @@ func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Res
...
+// declared by a workflow. Names are already lower-cased by actionlint since
+// GitHub Actions input names are case-insensitive.
+func workflowDispatchInputNames(workflow *actionlint.Workflow) []string {
+	var names []string
+	for _, event := range workflow.On {
+		dispatch, ok := event.(*actionlint.WorkflowDispatchEvent)
+		if !ok {
+			continue
+		}
+		for name := range dispatch.Inputs {
```
jmeridth: :robot: (medium) `type: choice` and `type: boolean` inputs get flagged as Failed here, but GitHub itself validates those at dispatch time (the UI restricts them and the REST API returns 422 for a value outside the declared set), which is exactly the "exit on expected values" sanitization the requirement's own recommendation describes. A workflow using a fixed-options choice input directly is platform-sanitized, so failing it is a false positive.

The AST already carries what you need: `actionlint.DispatchInput` has `.Type` (`WorkflowDispatchEventInputTypeChoice` / `...Boolean`) and `.Options`. This loop currently keeps only the map key, so the change is to collect the `*DispatchInput` values and skip Boolean, and Choice with a non-empty `Options` list, while continuing to flag String, Number, Environment, and untyped inputs. A test with one choice input and one string input in the same workflow would pin the distinction.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -898,3 +898,165 @@ func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Res
...
+// input prior to use in the pipeline.
+//
+// "Trusted collaborator input" here is workflow_dispatch inputs: values a
+// collaborator with write access types in when manually triggering a
+// workflow. Unlike OSPS-BR-01.01's fixed list of GitHub-controlled untrusted
+// metadata, workflow_dispatch input names are workflow-specific, so this
+// check derives the sensitive expressions from each workflow's own declared
+// inputs, then flags any of them interpolated directly into a run: step's
+// script body. An input assigned to a step-level env: var and referenced as
+// a shell variable is not flagged, matching the same convention used by
```
jmeridth: :robot: (medium) The env-indirection exemption is the right convention to follow, but as written it credits more than the safe form. Assigning an input to `env:` and reading it as a shell variable (`"$TARGET"`) is the recommended mitigation. Re-interpolating the env context back into the script body (`run: ./deploy.sh ${{ env.TARGET }}`) is the same injection with one extra hop, and we verified it returns Passed today. Full dataflow tracking is out of scope for a static check like this, so the practical ask is accuracy: narrow this comment (and the Passed message below) to say the check treats env assignment as the mitigation boundary and does not analyze `env.*` re-interpolation, so readers of the result know what was and was not proven. If you want to go one step further later, flagging `${{ env.X }}` inside `run:` bodies of jobs whose `env:` maps X from an input would close the gap, but that is follow-up territory.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -898,3 +898,165 @@ func DependenciesUseStandardizedTooling(payload data.Payload) (result gemara.Res
...
+//     into a run: step's script body.
+//   - NeedsReview: one or more workflow files could not be read (truncated
+//     by the API or unparseable), so an unsanitized input could be hiding
+//     there.
+//   - Passed: every declared workflow_dispatch input is either unused in
+//     run: steps or only referenced via a safer indirection.
+//   - NotApplicable: no workflow declares workflow_dispatch inputs, so there
+//     is no trusted collaborator input for a CI/CD pipeline to sanitize.
+func CicdSanitizesCollaboratorInput(payload data.Payload) (gemara.Result, string, gemara.ConfidenceLevel) {
+	var confidence gemara.ConfidenceLevel
```
jmeridth, line 952: :robot: (low, non-blocking) This declared-but-never-assigned variable means all four verdicts report confidence 0, which gemara renders as Undetermined. You inherited this faithfully from `checkAllWorkflows` above, so it is the local pattern, but the repo's newer step families set an explicit level on every return and it reads better in reports. If you want to adopt that here:

```suggestion
	confidence := gemara.Medium
```

and then return `gemara.High` for the NotApplicable branches and `gemara.Low` for NeedsReview in `evaluateCollaboratorInputSanitization` (which would need to return a confidence too). Also worth having one test assert the confidence value so it stays pinned. Entirely optional for this PR; the siblings share the pattern and a repo-wide cleanup may be the better vehicle.

evaluation_plans/osps/build_release/steps_test.go
```diff
@@ -1436,3 +1436,347 @@ func TestEnsureLatestReleaseHasChangelog(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			result, message := evaluateCollaboratorInputSanitization(tt.workflows)
+			t.Log(message)
+			assert.Equal(t, tt.wantResult, result)
+			assert.Contains(t, message, tt.wantContains)
+		})
+	}
+}
+
+// A Payload with no repository data configured surfaces GetWorkflowFiles'
```
jmeridth: :robot: (low, non-blocking) This comment argues for exactly the right behavior (surfacing "we could not check" rather than claiming the control does not apply), but the assertion below checks `gemara.NotApplicable`, which is the opposite of what the comment advocates. The behavior is inherited from `checkAllWorkflows`, so changing it here alone would diverge from the siblings; the quick fix for this PR is aligning the comment with what the test actually asserts, and the honest-verdict question (NeedsReview on retrieval errors, which matters more now that gemara halts on NotApplicable) can be a repo-wide follow-up covering all three steps that share the pattern.

jmeridth (changes requested):
:robot: (high) Welcome, and thank you for this contribution! Four-agent review per our tiered workflow (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5, Large tier for the 520-line diff); every finding below was verified by executing the real code, and the one high-severity finding was reproduced independently by two reviewers.

First, what you got right, because it is a lot, especially for a first PR here: the catalog entry is a verbatim port of the upstream OSPS Baseline text with the maturity level correctly translated (we checked against ossf/security-baseline three separate times), placing it only in the 2026-02 catalog is correct and the consistency tests agree, the doc comment nails the BR-01.01 vs BR-01.04 distinction that is the whole point of the control, the code mirrors the established evaluate* split with unreadable files tiered to NeedsReview, it collects every violation rather than stopping at the first (with a test proving it, and that is better than the sibling it was modeled on), the regex work is QuoteMeta-escaped, case-insensitive, and prefix-collision-tested, and 344 lines of table-driven tests for 162 lines of code is a great ratio. gofmt, vet, the full suite, and the race detector are all clean, and the branch merges onto current main without conflict.

What blocks the merge is one high and two mediums, each with a concrete fix in the line comments: the whole-expression anchoring lets every compound expression (the common `${{ inputs.x || 'default' }}` idiom, `format(...)`, ternaries, `toJSON(github.event.inputs)`) pass undetected; GitHub-validated `choice`/`boolean` inputs are flagged as Failed when the platform already enforces the sanitization the requirement asks for; and the env-indirection exemption credits `${{ env.X }}` re-interpolation that is equally injectable, so the doc comment and Passed message should state the actual boundary. Two low non-blocking notes accompany them (explicit confidence, and a test comment that argues against its own assertion).

For the maintainers rather than the author: BR-01.04 is Maturity Level 3 only while the example config and CI smoke tests pin Level 1, so no automated path currently executes this step; the README implemented-count convention is applied inconsistently across recent PRs and could use a ruling; the `run:`-only scope (shared with BR-01.01) misses the `actions/github-script` `with: script:` sink and could be a shared follow-up; and this file now contains a third copy of the workflow-walk loop, which is pre-existing debt, not this PR's.

Happy to re-review as soon as the fixes land; the regex change plus the choice/boolean exemption are each small, and the test rows to pin them are listed in the comments.

evaluation_plans/osps/build_release/steps.go
jmeridth: :robot: (informational) Verified fixed at 1cdd617 by execution: all four compound-expression cases now fail, the bare form still fails, and prefix-collision plus hyphenated-name safety held. The separate whole-object-dump pattern goes beyond what we asked for and is a genuinely good catch, since a bulk dump sidesteps the per-field exemptions. Resolving.

evaluation_plans/osps/build_release/steps.go
jmeridth: :robot: (informational) Verified fixed at 1cdd617: Boolean and option-constrained Choice inputs are exempt, and keeping Choice with no declared options flagged as a validation escape hatch is a subtlety our comment did not even ask for. Resolving.

evaluation_plans/osps/build_release/steps.go
jmeridth: :robot: (informational) Verified at 1cdd617: the doc comment and Passed message now state the env boundary and its limits precisely, and the dataflow follow-up framing is exactly right. Resolving.

evaluation_plans/osps/build_release/steps.go
jmeridth: :robot: (informational) Left as the sub-family convention, which this comment framed as entirely optional for this PR; a repo-wide confidence cleanup remains the better vehicle. Resolving as deferred.

evaluation_plans/osps/build_release/steps_test.go
jmeridth: :robot: (informational) Verified at 1cdd617: the comment now accurately describes the inherited behavior instead of contradicting the assertion. Resolving.

jmeridth (approved):
:robot: (low, non-blocking) Approving at 1cdd617, and for a first contribution this is genuinely excellent work. Every finding was re-verified by executing the real code: the boundary-guard pattern catches all the compound-expression forms while preserving the prefix-collision protection, the new whole-object-dump detection goes beyond what the review asked for, the platform-validated input exemption correctly keeps optionless Choice inputs flagged as an escape hatch, and the documentation now states exactly what the check does and does not analyze. 316 lines of new tests pin all of it, and the full suite passes with the race detector. The one deferred item (explicit confidence levels) is a repo-wide cleanup that should not fall on this PR. Welcome aboard, and thank you for the careful, thorough response to review.

PR #445, opened 2026-08-13
feat(OSPS-BR-02.02): associate release assets with release identifiers

### LINKED ISSUE #17: Implement OSPS-BR-02
The following test requirements are not yet implemented:

- [ ] When an official release is created, that release MUST be assigned a unique version identifier.
- [ ] When an official release is created, all assets within that release MUST be clearly associated with the release identifier or another unique identifier for the asset.

[OSPS-BR-02](https://baseline.openssf.org/versions/2025-02-25#osps-sa-0301:~:text=564%2C%20357%2D352-,OSPS%2DBR%2D02,-%2D%20All%20releases%20and)

jmeridth:
Closes #17

## What/Why

Implements OSPS-BR-02.02, the remaining half of the parent issue (BR-02.01 already landed): release assets must be clearly associated with the release identifier or another unique identifier. The new `ReleaseAssetsAssociatedWithRelease` step evaluates the release GitHub designates latest (its documented non-prerelease, non-draft definition, honoring an explicit maintainer designation), read from the GraphQL `latestRelease` already fetched for the changelog check, so both controls report on the same release at no extra API cost. Asset names must embed the release identifier (tag, v-toggled tag, or an unspaced release name) as a delimiter-bounded substring. Prereleases, drafts, sub-component release trains, and older releases do not drag the verdict down. Signature/checksum/attestation companions are recognized by the same classifiers BR-06.01 uses, and SBOMs by an extension set now shared with the QA-02.02 SBOM check, all associated through the artifact they accompany. Unrecognized identifiers surface as `NeedsReview` rather than `Failed` because the control permits "another unique identifier" the scanner cannot observe.

Degraded data cannot masquerade as a clean verdict: the payload loader's GraphQL soft-failure path is now recorded as `Payload.GraphqlPartialData` (following the `InsightsError`/`RulesetsObserved` observability convention), and an empty `latestRelease` on partial data degrades to `NeedsReview` at Low instead of reading as a repository with no releases. Every return carries an explicit confidence level.

## Proof it works

- 18-case table test asserting result, message, and confidence for every branch: unobservable data, partial GraphQL resolution, no designated latest (prerelease-only repos are NotApplicable), a row proving the REST release list is not consulted, tag/v-toggled/release-name association, case-insensitivity, companion exemptions, delimiter-bounded matching in both directions (v1 does not match v10), message capping at five with an overflow note, single-character tag/name overmatch guards, blank-named assets, and companion-only releases.
- Data-layer tests assert the partial-resolution flag on all three GraphQL fetch paths (clean, soft failure, hard failure).
- Helper tests for identifier-candidate derivation and companion classification, including delegation to the shared signature/checksum classifiers and the shared SBOM extension set.
- `go test ./...` across all packages and `go test -race` on the touched packages pass locally.

## Risk + AI role

Low: a new single-step check plus wiring, a shared SBOM extension set consumed by QA-02.02, and an additive observability flag in the payload loader; clean-data verdicts of existing controls are unchanged. AI-generated end to end with maintainer-approved design; the interpretation (naming heuristic with `NeedsReview`, no `Failed` path, and latest-release scoping) was an explicit maintainer decision over structural-only or fail-on-missing readings.

## Review focus

- The evaluation-time scoping: only GitHub's designated latest release is judged; no designated latest (including prerelease-only repos) or a latest release with no assets is `NotApplicable` at High confidence, while an empty `latestRelease` on partially resolved GraphQL data is `NeedsReview` at Low.
- The `GraphqlPartialData` flag is deliberately coarse: any field-level GraphQL error sets it, trading message precision for a one-field data-layer change.


jmeridth:
🤖 e922b1f rewords this PR's new doc comment to describe behavior instead of the catalog ID, per #448 category 1. The ID linkage stays in the dispatch map entry.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,151 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+	if releaseAssetCompanionNames[lowerName] {
+		return true
+	}
+	for _, suffix := range releaseAssetCompanionSuffixes {
+		if strings.HasSuffix(lowerName, suffix) {
+			return true
+		}
+	}
+	return false
+}
```
eddie-knight, line 541: 🤖 **Major** — This is now the repo's third independent companion/signature classifier, and the narrowest of the three. The same file already recognizes `.sign`, `.gpg`, `.pgp`, `.sigstore`, `.sigstore.json` (`signatureAssetKinds`, steps.go:849) and `.sha256sum`/`.sha512sum` plus substring `checksums` manifests (`isHashManifest`, steps.go:879 — which also catches bare `SHASUMS`-style names this exact-match map misses); `quality/steps.go:654` and `:495` additionally cover `.cert`/`.crt`/`.p7s`/`.sha`/`.sha224`/`.sha384` and bare `.spdx`/`.cdx`/`.cdx.xml`.

Concrete effect: a `v1.2.3` release shipping `mytool-1.2.3.tar.gz` + `mytool-1.2.3.tar.gz.gpg` (or `SHASUMS`, or `bom.spdx`) reports NeedsReview naming the companion as unassociated, while BR-06.01 in this same file classifies the identical name as a signature/manifest — two controls disagreeing about one asset.

Suggest delegating: `signatureAssetKind(lowerName) != "" || isHashManifest(lowerName)` plus the SBOM/doc additions kept locally (quality's classifiers are unexported; consolidating all three into `reusable_steps` would be the complete fix). Small extra while you're here: the exact-match `license`/`readme` keys miss the common `LICENSE-MIT`/`LICENSE-APACHE` convention — though a naive prefix match would also exempt a real artifact named e.g. `license-manager.zip`, so that one's a judgment call.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,151 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+	tag := strings.ToLower(strings.TrimSpace(release.TagName))
+	if tag != "" {
+		candidates = append(candidates, tag)
+		if trimmed := strings.TrimPrefix(tag, "v"); trimmed != tag && len(trimmed) >= 2 {
+			candidates = append(candidates, trimmed)
+		}
+	}
+	if name := strings.ToLower(strings.TrimSpace(release.Name)); name != "" && !strings.Contains(name, " ") {
+		candidates = append(candidates, name)
+	}
```
eddie-knight: 🤖 **Minor** — The two-character overmatch guard only protects the v-stripped tag; the raw tag and the release name are appended unguarded. With `TagName: "1"`, every asset whose name contains a `1` counts as associated (`tool-build-7781.zip` passes), which is exactly the overmatch the comment above says this guard prevents. Same floor on all three candidates:

```suggestion
	if len(tag) >= 2 {
		candidates = append(candidates, tag)
		if trimmed := strings.TrimPrefix(tag, "v"); trimmed != tag && len(trimmed) >= 2 {
			candidates = append(candidates, trimmed)
		}
	}
	if name := strings.ToLower(strings.TrimSpace(release.Name)); len(name) >= 2 && !strings.Contains(name, " ") {
		candidates = append(candidates, name)
	}
```

No existing test breaks (a single-char tag/name has no case yet — worth adding one); the doc comment above wants the matching tweak.

eddie-knight (commented):
🤖 Multi-model review — three independent reviewers (Haiku, Sonnet, Opus) over a local worktree at head 591cdbb, every finding re-verified against source by the orchestrator before posting. Requested focus: idiomatic, DRY, functionally complete.

**Focus answers**
1. **Idiomatic** — yes: step signature, per-branch confidence, `HasPublishedRelease` gating, and table-driven tests all match sibling steps. One structural break, covered inline.
2. **DRY** — no: `isReleaseAssetCompanion` is the repo's third independent companion/signature classifier and the narrowest of the three, and `releaseAssetLabel` copies `quality`'s `releaseLabel`. Details inline.
3. **Functionally complete** — yes for every branch the PR body describes (all present and tested, overflow arithmetic and draft exclusion verified), but the documented companion exemption doesn't fully hold (`.gpg`, `SHASUMS`, bare `.spdx` still get flagged) and the single-digit overmatch guard covers only one of the three identifier candidates. Both inline.

**Blocking-ish:** the classifier consolidation (major) and the candidate length guard (minor) — everything else is take-or-leave.

**Checked and NOT problems** (recorded so they don't get re-raised):
- `evaluation-plans.go` wiring: only the OSPS-BR-02.02 entry changed; BR-02.01 and BR-06.01 code paths untouched.
- Overflow message arithmetic (`and %d more`) computes the count before truncation — correct.
- One reviewer's claim that the companion list is "comprehensive" was refuted by direct read of `signatureAssetKinds`/`isHashManifest` (same file) and `quality`'s classifiers; the gap is the inline major.
- `go build` / `go vet` / `go test ./evaluation_plans/...` pass at head.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,151 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+	for _, candidate := range candidates {
+		if strings.Contains(lowerName, candidate) {
+			return true
+		}
+	}
+	return false
+}
+
+// releaseAssetLabel names a release in evidence messages, preferring the tag.
+func releaseAssetLabel(release data.ReleaseData) string {
```
eddie-knight: 🤖 **Minor** — Verbatim copy of `releaseLabel` (`quality/steps.go:484-492`) apart from the fallback string; both take `data.ReleaseData` and both packages already import `reusable_steps`. Worth moving one copy there and deleting the other — same root cause as the classifier comment above.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,151 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+	totalAssets := 0
+	checkedAssets := 0
+	var unassociated []string
+	for _, release := range payload.Releases {
+		if release.Draft {
+			continue
+		}
+		candidates := releaseIdentifierCandidates(release)
+		for _, asset := range release.Assets {
+			totalAssets++
```
eddie-knight: 🤖 **Nit** — a blank/whitespace-only asset name increments `totalAssets` but is skipped, so a release containing only such assets reports "only companion files (checksums, signatures, documentation)". Unlikely input from GitHub, but skipping before counting makes the message honest.

evaluation_plans/osps/build_release/steps_test.go
```diff
@@ -2070,3 +2070,222 @@ func TestCicdSanitizesCollaboratorInputNoWorkflows(t *testing.T) {
...
+	}
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			assert.Equal(t, testCase.want, releaseIdentifierCandidates(testCase.release), testCase.name)
+		})
+	}
+}
+
+func TestIsReleaseAssetCompanion(t *testing.T) {
+	cases := map[string]bool{
```
eddie-knight, line 2283: 🤖 **Minor** — Every `true` case here is a literal entry of the two new lists, so this test can't detect the coverage gap flagged on `isReleaseAssetCompanion` — once that delegates to the existing classifiers, add expected-true cases for `.gpg`, `.sigstore`, bare `.spdx`, and `SHASUMS`.

evaluation_plans/osps/build_release/steps_test.go
```diff
@@ -2070,3 +2070,222 @@ func TestCicdSanitizesCollaboratorInputNoWorkflows(t *testing.T) {
...
+				{TagName: "v-next", Draft: true, Assets: []data.ReleaseAsset{{Name: "unversioned.zip"}}},
+				assetRelease("v1.2.3", "", "mytool_1.2.3.zip"),
+			}}},
+			wantResult:     gemara.Passed,
+			wantMsgPart:    "All release assets embed a release identifier",
+			wantConfidence: gemara.Medium,
+		},
+	}
+
+	for _, testCase := range tests {
```
eddie-knight: 🤖 **Nit** — `testCase` introduces a third loop-variable convention in this file (11 existing loops use `tt`, 4 use `tc`) — here and at line 2265.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2. Delegated to signatureAssetKind and isHashManifest as suggested, with the local lists reduced to what those do not cover: .pem (not in signatureAssetKinds), SBOM suffixes including bare .spdx/.cdx/.cdx.xml, documentation names, and md5sums (the one manifest name the shared markers miss). On the license judgment call, went with exact license-mit and license-apache entries rather than a prefix, and license-manager.zip has a test pinning it as a real artifact. Consolidating all three classifiers into reusable_steps needs quality's unexported classifiers moved too, so I added it to the #448 cleanup checklist rather than growing this PR.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2 with your suggestion as written, plus cases for a single-character tag, a single-character release name, and a behavioral case pinning that TagName 1 no longer associates tool-build-7781.zip.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2. Moved to reusable_steps.ReleaseLabel and deleted both local copies; quality's call site now uses the shared helper and the fallback string is the more descriptive of the two.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2. Blank names are skipped before counting, and a fixture pins a blank-only release reporting no attached assets instead of the companion message.

evaluation_plans/osps/build_release/steps_test.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2. Added expected-true cases for .gpg, .sigstore, bare .spdx, .cdx.xml, and a version-prefixed checksums manifest, all of which failed against the old lists, plus md5sums as a regression guard since the shared markers do not cover it.

evaluation_plans/osps/build_release/steps_test.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/825d4b77dfc091559361f346de433a5b5f609dc2. Normalized the new loops to tt, and while doing so found two older loops in this file also using testCase, so those are converted as well; tt is now the only convention alongside the four existing tc loops.

jmeridth:
🤖 0d2130b tightens the evaluation scope to the most recent published release, per the option discussed. The requirement is judged against the release process as it stands at evaluation time, so older releases that predate a naming convention no longer drag the verdict down.

Semantics now: the first non-draft entry in the newest-first release list is the one evaluated; drafts are skipped as before. A most recent release with no assets is NotApplicable even when older releases have assets. Messages name the evaluated release. Two red-first matrix rows pin the scoping in both directions: a clean current release passes despite a dirty older one, and a dirty current release needs review despite clean older ones.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,163 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+func ReleaseAssetsAssociatedWithRelease(payload data.Payload) (gemara.Result, string, gemara.ConfidenceLevel) {
+	released, observable := reusable_steps.HasPublishedRelease(payload)
+	if !observable {
+		return gemara.NeedsReview, "Release data is unavailable; manually review whether release assets are associated with the release identifier", gemara.Low
+	}
+	if !released {
+		return gemara.NotApplicable, "No published releases found; the asset-identifier requirement does not apply", gemara.High
+	}
+
+	// The REST release list is newest-first, so the first non-draft entry is
```
eddie-knight: 🤖 **Major** — this comment states an ordering guarantee GitHub does not give, and the rule it justifies selects the wrong release on real repos.

GitHub documents no sort order for `GET /repos/{owner}/{repo}/releases`. It does document the definition of the latest release, which is stricter than this rule: *"the most recent non-prerelease, non-draft release."* This excludes drafts only. `data.ReleaseData` (data/rest-data.go:56-63) has no `Prerelease`, `PublishedAt`, or `CreatedAt`, so the step can neither exclude prereleases nor sort.

Checked live:

| repo | selected here | actual latest |
|---|---|---|
| `rancher/rancher` | `v2.14.5-alpha1` — prerelease, 0 assets | `v2.15.0`, 17 assets |
| `grafana/loki` | `operator/v0.11.0` — not a prerelease, 0 assets | `v3.7.6`, 47 assets |
| `istio/istio` | `1.31.0-rc.0` — prerelease | `1.30.3` |

rancher and loki return `NotApplicable` ("The most recent release has no attached assets"). istio judges the RC, whose assets embed `1.31.0-rc.0` and pass; a badly-named stable release would be masked.

`grafana/loki` rules out "skip prereleases" as a complete fix: `operator/v0.11.0` is a sub-component release train, not a prerelease.

`EnsureLatestReleaseHasChangelog` (steps.go:846) reads `payload.Repository.LatestRelease` — the GraphQL `latestRelease`, which does follow GitHub's definition. Verified live: GraphQL returns `1.30.3` for istio where the REST list's first entry is `1.31.0-rc.0`. The two controls would report on different releases in one run.

Fix options: wire up `data/graphql-data.go:81-91`, which already queries `releases(first: 1, orderBy: {field: CREATED_AT, direction: DESC})` with tag, name, and asset names and has no consumer; or add `Prerelease bool` + `PublishedAt` to `ReleaseData` and select explicitly. If list-order-first is intended as a deliberate approximation, state that and its failure modes here rather than asserting the ordering.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,163 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
 	return gemara.Passed, "All releases found have a unique name", confidence
 }
 
+// releaseAssetCompanionSuffixes identify signature, checksum, attestation, and
+// SBOM companion files. A companion is associated with the artifact it
+// accompanies rather than carrying the release identifier itself, so it is
+// exempt from the name-association expectation.
+var releaseAssetCompanionSuffixes = []string{
+	".pem",
+	".sbom", ".spdx", ".spdx.json", ".cdx", ".cdx.json", ".cdx.xml",
```
eddie-knight: 🤖 **Minor** — this SBOM list is a partial copy of QA-02.02's, so the two controls classify the same file differently. `quality/steps.go:482-484` lists `.spdx.yaml .spdx.yml .spdx.rdf .spdx.xml`; none are here, and none suffix-match `.spdx`. Verified by executing the predicate: `bom.spdx.yaml`, `bom.spdx.xml`, and `bom.spdx.rdf` all return `companion=false`, so an SBOM to QA-02.02 is a checkable artifact here and lands in `NeedsReview`.

Fix: move `isSBOMAsset`'s extension check to `reusable_steps` and delegate, as 825d4b7 did for the signature classifiers.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,163 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+
+// releaseIdentifierCandidates returns the lowercased identifier strings an
+// asset name may embed to associate itself with the release: the tag, the tag
+// without its leading "v", and the release name when it contains no spaces
+// (asset file names cannot contain a spaced title). Every candidate must be at
+// least two characters so a single digit cannot match almost any name.
+func releaseIdentifierCandidates(release data.ReleaseData) []string {
+	var candidates []string
+	tag := strings.ToLower(strings.TrimSpace(release.TagName))
+	if len(tag) >= 2 {
```
eddie-knight, line 553: 🤖 **Minor** — the two-character floor drops the trimmed candidate but not the raw tag. Tag `v1` yields candidate `"v1"`, and `assetNameContainsAny` is a bare `strings.Contains`, so `myapp-v10.zip` reports `Passed`. Verified by execution: `candidates=[v1]`, `contains("myapp-v10.zip")=true`.

Fix: require the candidate to be delimiter-bounded, as `quality/steps.go:536-540` does for `sbomTokens`.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,163 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+var releaseAssetCompanionSuffixes = []string{
+	".pem",
+	".sbom", ".spdx", ".spdx.json", ".cdx", ".cdx.json", ".cdx.xml",
+}
+
+// releaseAssetCompanionNames are exact (lowercased) asset names that accompany
+// a release without identifying a specific artifact: standard documentation
+// files, common split-license names, and the one manifest name the shared
+// checksum classifier's markers do not cover. Names are matched exactly rather
+// than by prefix so a real artifact such as license-manager.zip is not exempted.
```
eddie-knight: 🤖 **Nit** — after 825d4b7's delegation the exact-match rationale no longer holds. `hashManifestMarkers` (steps.go:875) are substring-matched (steps.go:897-900), so a real artifact can be exempted through that path: `checksums-generator-1.0.zip` returns `companion=true`. Verified by execution. The `license-manager.zip` example still holds for the map itself; the comment should note the delegated markers' substring semantics.

eddie-knight (commented):
🤖 Multi-model panel review (haiku / sonnet / opus) at `0d2130b`. Four findings, all verified against source, the GitHub REST docs, and the live API.

**Major — `steps.go:604`, release selection.** The comment asserts an ordering GitHub does not document, and "first non-draft" is not GitHub's definition of the latest release ("most recent non-prerelease, non-draft"). `data.ReleaseData` has no `Prerelease` or timestamp field, so list order is the only signal available. Checked live:

| repo | selected here | actual latest |
|---|---|---|
| `rancher/rancher` | `v2.14.5-alpha1` — prerelease, 0 assets | `v2.15.0`, 17 assets |
| `grafana/loki` | `operator/v0.11.0` — not a prerelease, 0 assets | `v3.7.6`, 47 assets |
| `istio/istio` | `1.31.0-rc.0` — prerelease | `1.30.3` |

rancher and loki return `NotApplicable`; the control is skipped on repos that do publish assets. `grafana/loki` rules out "skip prereleases" as a complete fix — `operator/v0.11.0` is a sub-component train, not a prerelease. Detail and fix options inline.

**Minor — `steps.go:504`.** SBOM suffix list diverges from QA-02.02's; `bom.spdx.yaml` is an SBOM there and a checkable artifact here.

**Minor — `steps.go:548`.** Tag `v1` yields candidate `"v1"`, so `myapp-v10.zip` passes on a bare `strings.Contains`.

**Nit — `steps.go:511`.** The exact-match rationale in the comment is defeated by the delegated substring markers.

Also relevant to the major: `data/graphql-data.go:81-91` already queries `releases(first: 1, orderBy: {field: CREATED_AT, direction: DESC})` with tag, name, and asset names, with an explicit ordering, and has no consumer.

evaluation_plans/osps/build_release/steps.go
eddie-knight: Confirmed this manually and also had a Fable instance give it another pass with a real execution

---

🤖 Confirmed end-to-end, not just by reasoning over API data. Built this branch at `0d2130b` as a plugin, ran `pvtr run` against `rancher/rancher` through the full data-loading path:

```
OSPS-BR-02.02
  result: Not Applicable
  message: The most recent release has no attached assets;
           the asset-identifier requirement does not apply
```

Corroboration from the same scan: QA-02.02 reports hundreds of rancher releases that publish artifacts (`v2.15.0`, `v2.14.3`, `v2.13.3`, …), while BR-02.02 in the same run sees none. The selection lands on `v2.14.5-alpha1` — prerelease, 0 assets — where `/releases/latest` is `v2.15.0` with 17.

One setup note for anyone reproducing: the released `pvtr` 0.22.0 hits a gob protocol mismatch against this branch's SDK pin (v1.32.2), so this must be run with `v0.23.0` or later.

jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/1dc309070d294ba83f0a9455a52a7d06e23dd3c6. Went with the GraphQL route, but via the latestRelease field rather than the releases(first: 1, CREATED_AT DESC) node: created-at ordering would still pick rancher's alpha, while latestRelease follows GitHub's documented definition and honors a maintainer's explicit latest designation, which the loki operator-train case shows is the only authoritative source. It also rides the query the changelog check already runs, so both controls now report on the same release with no extra API calls. An empty latestRelease (prerelease-only repos included) is NotApplicable, and a regression row proves the REST list is no longer consulted. Thank you for running this end to end; the rancher scan made the failure unambiguous.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/1dc309070d294ba83f0a9455a52a7d06e23dd3c6. The extension set moved to reusable_steps.HasSBOMExtension and both controls delegate to it; bom.spdx.yaml, bom.spdx.xml, and bom.spdx.rdf are red-first companion cases now.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/1dc309070d294ba83f0a9455a52a7d06e23dd3c6. Candidates are now delimiter-bounded on both sides (non-alphanumeric or string edge), with myapp-v10.zip rejected and myapp-v1.zip accepted as red-first rows. While in there, an unprefixed tag also gains a v-prefixed candidate so tag 1.2.3 matches mytool-v1.2.3.zip.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/1dc309070d294ba83f0a9455a52a7d06e23dd3c6. The comment now states that the shared checksum markers are substring-matched and names the checksums-generator-1.0.zip consequence, which is also pinned as a documented-behavior test case.

evaluation_plans/osps/build_release/steps.go
```diff
@@ -494,6 +495,184 @@ func ReleaseHasUniqueIdentifier(payload data.Payload) (result gemara.Result, mes
...
+//   - NeedsReview: release data unobservable, at least one non-companion asset
+//     of the latest release does not embed a release identifier, or that
+//     release publishes only companion files.
+//   - NotApplicable: no release is designated latest (including
+//     prerelease-only repositories), or the latest release publishes no
+//     assets.
+//   - Passed: every non-companion asset of the latest release embeds a
+//     release identifier.
+func ReleaseAssetsAssociatedWithRelease(payload data.Payload) (gemara.Result, string, gemara.ConfidenceLevel) {
+	if payload.GraphqlRepoData == nil {
```
eddie-knight, line 623: 🤖 **Minor** — this guard is unreachable in production, and the degraded-data case it was protecting now returns a high-confidence `NotApplicable`.

`NewPayload` returns an error if the GraphQL fetch hard-fails, and dereferences `graphql.Repository.DefaultBranchRef.Name` unconditionally at data/payload.go:130 — so a `Payload` with a nil `GraphqlRepoData` never exists outside tests.

The reachable degraded case is the soft-failure path in `getGraphqlRepoData`: when the query returns an errors array but `Repository.Name` resolved, it logs a warning and proceeds on partial data where the failed fields, in that function's own words, "decode to their zero value." If `latestRelease` is among them, `latest.TagName` and `latest.Name` are both empty and this falls through to the `NotApplicable` at :631. Verified by execution:

```
partial-graphql (latestRelease zeroed) -> Not Applicable (High)
no-releases                            -> Not Applicable (High)
```

Same result, same confidence — no way to tell a fetch problem from a repository that genuinely has no designated latest release. That loses the property the PR description advertises, "unobservable release data degrades to NeedsReview, with explicit confidence on every return," which did hold at `0d2130b` through `HasPublishedRelease`.

No suggestion block here because the obvious one-liner is wrong, and it may save you the detour: gating on `reusable_steps.HasPublishedRelease` ahead of the `latestRelease` read fails 15 cases in `TestReleaseAssetsAssociatedWithRelease`, because `latestReleasePayload` builds payloads with no `RestData` — and it would recouple a GraphQL-sourced check to REST, which is what this commit deliberately moved away from.

There is also no clean in-band signal to distinguish the two: `Repository.Name` is non-empty in both, since a populated name is the precondition for the soft-failure path returning at all. So the options are roughly: have `getGraphqlRepoData` record which fields failed to resolve so steps can consult it; cross-check REST only inside the no-latest branch and only when `RestData` is present; or accept the ambiguity and drop that `NotApplicable` from `High` to `Medium` confidence. The last is a one-word change and honest about what the scanner actually knows.

eddie-knight (commented):
🤖 Re-reviewed at `1dc3090`. One regression inline.

evaluation_plans/osps/build_release/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/c95dbea99d734c6479cfe1f35bd6cb63fbd30534 with the first option in coarse form, and thank you for pre-empting the HasPublishedRelease detour, that is exactly where I would have gone first. getGraphqlRepoData now reports the soft-failure path and NewPayload records it as Payload.GraphqlPartialData, following the same observability convention as InsightsError and RulesetsObserved. The step consults it only inside the empty-latest branch: partial data degrades to NeedsReview at Low, a cleanly resolved absence keeps NotApplicable at High, so genuinely release-less repos are not penalized. The flag is deliberately coarse (any field-level error sets it); per-field tracking felt like overkill for one consumer, and the flag is now available to the other steps with the same zero-value blind spot. The unreachable nil guard stays as a panic guard for zero-value payloads. Data-layer tests assert the flag on all three fetch paths and a step row pins the empty-plus-partial case.

eddie-knight (approved):

PR #446, opened 2026-08-13
feat(OSPS-GV-04.01): require review before escalated permissions

### LINKED ISSUE #21: Implement OSPS-GV-04
The following test requirements are not yet implemented:

- [ ] While active, the project documentation MUST have a policy that code contributors are reviewed prior to granting escalated permissions to sensitive resources.

[OSPS-GV-04](https://baseline.openssf.org/versions/2025-02-25#osps-sa-0301:~:text=OC%3A%204.1.2-,OSPS%2DGV%2D04,-%2D%20The%20project%20documentation)

jmeridth:
Closes #21

## What/Why

Implements OSPS-GV-04.01: project documentation must have a policy that code collaborators are reviewed prior to receiving escalated permissions to sensitive resources. The new `HasEscalatedPermissionsReviewPolicy` step scans repository documentation prose per markdown section for escalation vocabulary (maintainer, committer, write access, elevated permissions) paired with review vocabulary (reviewed, approved, nominated, vote): an uncontradicted requirement passes at Medium confidence; a pairing without obligation language, a negated requirement, a declared Security Insights `governance` URL, or a bare governance file all route to `NeedsReview` with distinct messages; and an observed absence of every signal fails, mirroring the GV-03.02 terminal. Code fences are stripped before analysis, vocabulary must co-occur within one section, negated or optional phrasing is never credited (the lesson from the VM-05 prose checks), and unreadable documentation is never reported as a violation.

## Proof it works

- 9-case step table asserting result, message, and confidence for every branch: non-code gating, clear policy, negated policy, mention without obligation, split-section isolation, fenced-code exclusion, governance-file fallback, unreadable docs, and observed absence.
- Security Insights fallback test plus a 7-case section-classifier table.
- `go test ./...` across all packages and `go test -race` on the governance package pass locally.

## Risk + AI role

Low: a new single-step check plus wiring; no existing control logic modified. AI-generated end to end with maintainer-approved design; allowing a `Passed` path guarded by negation screening (rather than a never-pass check) was an explicit maintainer decision.

## Review focus

- Vocabulary breadth: are the escalation and obligation patterns wide enough for real governance docs without inviting false passes?
- Whether `Passed` at Medium from a prose heuristic is the right ceiling for this MUST control.

Note: this will take a one-line mechanical README conflict against #445 for whichever merges second (both bump the implemented-controls count from 42).

### DIFF
--- a/evaluation_plans/osps/governance/steps.go
+++ b/evaluation_plans/osps/governance/steps.go
@@ -1,6 +1,7 @@
 package governance
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/gemaraproj/go-gemara"
@@ -145,4 +146,138 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
 	}
 
 	return gemara.Failed, "No contributor guide documenting requirements for acceptable contributions found in Security Insights data or repository files", gemara.Medium
+}
+
+// Vocabulary for OSPS-GV-04.01. All patterns are word-boundary anchored so
+// unrelated words cannot match as substrings.
+var (
+	// escalationVocabPattern recognizes text about collaborator roles and the
+	// permissions those roles carry.
+	escalationVocabPattern = regexp.MustCompile(`(?i)\b(?:maintainer|committer|collaborator|commit access|write access|push access|merge rights|admin(?:istrator)?\s+(?:role|rights|access)|elevated\s+(?:permissions?|privileges?|access)|escalated\s+permissions?|owner\s+role)\b`)
+
+	// reviewVocabPattern recognizes text about a person being reviewed,
+	// approved, nominated, or vetted.
+	reviewVocabPattern = regexp.MustCompile(`(?i)\b(?:review(?:ed)?|approv(?:e|ed|al)|nominat(?:e|ed|ion)|vote(?:d)?|consensus|sponsor(?:ed)?|vett(?:ed|ing))\b`)
+
+	// escalationObligationPattern recognizes language that makes the review a
+	// requirement rather than a description.
+	escalationObligationPattern = regexp.MustCompile(`(?i)\b(?:must|shall|required?|requires|only\s+(?:after|by|through|via)|needs?\s+to\s+be\s+approved)\b`)
+
+	// escalationNegationPattern recognizes language that disclaims or weakens
+	// the requirement within the same section, so contradictory prose is not
+	// credited as a definitive policy (the lesson from the VM-05 prose checks).
+	escalationNegationPattern = regexp.MustCompile(`(?i)\b(?:not\s+required|no\s+(?:formal\s+)?review|optional|informal(?:ly)?|do(?:es)?\s+not\s+require|without\s+(?:a\s+)?review)\b`)
+
+	// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
+	// example snippets cannot satisfy or negate the policy vocabulary.
+	fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
+)
+
+// escalationPolicySections splits a documentation file into markdown-heading
+// sections with fenced code blocks removed, so vocabulary co-occurrence is
+// judged within one topical section rather than across an entire file.
+func escalationPolicySections(content string) []string {
+	cleaned := fencedCodeBlockPattern.ReplaceAllString(content, "")
+	var sections []string
+	var current []string
+	flush := func() {
+		if len(current) > 0 {
+			sections = append(sections, strings.Join(current, "\n"))
+			current = nil
+		}
+	}
+	for _, line := range strings.Split(cleaned, "\n") {
+		if strings.HasPrefix(strings.TrimSpace(line), "#") {
+			flush()
+		}
+		current = append(current, line)
+	}
+	flush()
+	return sections
+}
+
+// classifyEscalationSection reports whether a section mentions collaborator
+// escalation together with review (mention), and whether it additionally states
+// the review as an uncontradicted requirement (policy).
+func classifyEscalationSection(section string) (mention bool, policy bool) {
+	if !escalationVocabPattern.MatchString(section) || !reviewVocabPattern.MatchString(section) {
+		return false, false
+	}
+	if !escalationObligationPattern.MatchString(section) {
+		return true, false
+	}
+	if escalationNegationPattern.MatchString(section) {
+		return true, false
+	}
+	return true, true
+}
+
+// HasEscalatedPermissionsReviewPolicy implements OSPS-GV-04.01: while active,
+// the project documentation MUST have a policy that code collaborators are
+// reviewed prior to granting escalated permissions to sensitive resources.
+//
+// This is distinct from GV-03.02 (requirements for acceptable contributions):
+// the subject here is the person receiving elevated access, not the change
+// being merged. Security Insights has no escalation-policy field, so evidence
+// is layered: repository documentation prose is scanned for a section that
+// pairs escalation vocabulary with review vocabulary; an uncontradicted
+// requirement Passes at Medium (heuristic prose match), a weaker pairing or a
+// declared Security Insights governance URL or a governance file needs human
+// review, and an observed absence of every signal Fails, mirroring GV-03.02's
+// terminal. Unreadable documentation is never reported as a violation.
+func HasEscalatedPermissionsReviewPolicy(payload data.Payload) (result gemara.Result, message string, confidence gemara.ConfidenceLevel) {
+	if !payload.IsCodeRepo {
+		return gemara.NotApplicable, "Repository contains no code - skipping escalated-permissions policy check", gemara.High
+	}
+
+	files, docsErr := payload.GetDocumentationFiles()
+	policyFile := ""
+	mentionFile := ""
+	for _, file := range files {
+		for _, section := range escalationPolicySections(file.Content) {
+			mention, policy := classifyEscalationSection(section)
+			if policy && policyFile == "" {
+				policyFile = file.Path
+			}
+			if mention && mentionFile == "" {
+				mentionFile = file.Path
+			}
+		}
+	}
+
+	if policyFile != "" {
+		return gemara.Passed, "Documentation requires code collaborators to be reviewed before receiving escalated permissions (" + policyFile + ")", gemara.Medium
+	}
+	if mentionFile != "" {
+		return gemara.NeedsReview, "Documentation discusses collaborator roles and review (" + mentionFile + ") but does not state an uncontradicted requirement that collaborators are reviewed before escalated permissions are granted; manual review required", gemara.Medium
+	}
+	if payload.RestData != nil && payload.Insights.Repository != nil &&
+		payload.Insights.Repository.Documentation != nil &&
+		payload.Insights.Repository.Documentation.Governance != nil {
+		return gemara.NeedsReview, "Governance documentation is declared in Security Insights; confirm it requires review of code collaborators before escalated permissions are granted", gemara.Medium
+	}
+	if file := governanceFilePresent(files); file != "" {
+		return gemara.NeedsReview, "A governance document (" + file + ") is present but no escalation-review policy language was recognized; manual review required", gemara.Low
+	}
+	if docsErr != nil {
+		return gemara.NeedsReview, "Repository documentation could not be fully read; manually review whether a collaborator escalation-review policy is documented", gemara.Low
+	}
+
+	return gemara.Failed, "No policy requiring review of code collaborators before granting escalated permissions was found in repository documentation or Security Insights data", gemara.Medium
+}
+
+// governanceFilePresent returns the path of a conventional governance or
+// maintainers document among the fetched documentation files, or "".
+func governanceFilePresent(files []data.DocumentationFile) string {
+	for _, file := range files {
+		base := strings.ToLower(file.Path)
+		if slash := strings.LastIndex(base, "/"); slash >= 0 {
+			base = base[slash+1:]
+		}
+		switch base {
+		case "governance.md", "governance", "maintainers.md", "maintainers":
+			return file.Path
+		}
+	}
+	return ""
 }
--- a/evaluation_plans/osps/governance/steps.go
+++ b/evaluation_plans/osps/governance/steps.go
@@ -148,31 +148,136 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
 	return gemara.Failed, "No contributor guide documenting requirements for acceptable contributions found in Security Insights data or repository files", gemara.Medium
 }
 
-// Vocabulary for detecting an escalated-permissions review policy. All
-// patterns are word-boundary anchored so unrelated words cannot match as
-// substrings.
+// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
+// example snippets cannot satisfy or negate the policy vocabulary.
+var fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
+
+// escalationTokens lowercases prose and strips punctuation. No stemming is
+// attempted: vocabulary entries are written as stems and compared by prefix,
+// which handles inflection by construction rather than by guessing suffixes.
+func escalationTokens(text string) []string {
+	fields := strings.Fields(text)
+	tokens := make([]string, 0, len(fields))
+	for _, field := range fields {
+		token := strings.ToLower(strings.Trim(field, ".,;:!?()[]{}\"'`"))
+		if token != "" {
+			tokens = append(tokens, token)
+		}
+	}
+	return tokens
+}
+
+// escalationTokenMatches compares one prose token against one vocabulary token.
+// A vocabulary entry of four or more characters is treated as a stem, so
+// "approv" covers approve/approved/approval/approves/approving and "maintainer"
+// covers the plural, without enumerating forms. Short entries such as "to" must
+// match exactly, so they cannot match inside unrelated words.
+func escalationTokenMatches(token, vocabToken string) bool {
+	if len(vocabToken) >= 4 {
+		return strings.HasPrefix(token, vocabToken)
+	}
+	return token == vocabToken
+}
+
+// escalationVocabulary is a set of phrases, each one or more consecutive
+// normalized tokens. Extending a vocabulary means adding a string.
+type escalationVocabulary struct {
+	name    string
+	phrases [][]string
+}
+
+func newEscalationVocabulary(name string, entries ...string) escalationVocabulary {
+	vocabulary := escalationVocabulary{name: name}
+	for _, entry := range entries {
+		vocabulary.phrases = append(vocabulary.phrases, escalationTokens(entry))
+	}
+	return vocabulary
+}
+
+// find returns the index of the earliest matching phrase, or -1. It returns a
+// position rather than a bool because the control requires review *prior to*
+// escalation, and an ordering relation cannot be expressed with a bool.
+func (v escalationVocabulary) find(tokens []string) int {
+	earliest := -1
+	for _, phrase := range v.phrases {
+		if len(phrase) == 0 {
+			continue
+		}
+		for i := 0; i+len(phrase) <= len(tokens); i++ {
+			matched := true
+			for j, want := range phrase {
+				if !escalationTokenMatches(tokens[i+j], want) {
+					matched = false
+					break
+				}
+			}
+			if matched && (earliest == -1 || i < earliest) {
+				earliest = i
+			}
+		}
+	}
+	return earliest
+}
+
+func (v escalationVocabulary) has(tokens []string) bool { return v.find(tokens) >= 0 }
+
+// Vocabulary for the escalated-permissions review policy. Note the distinction
+// a single pattern cannot draw: a role noun ("maintainer") is not itself
+// evidence of escalation, while an access object ("write access") or a granting
+// verb applied to a role is. That separation is what keeps prose about
+// reviewing *changes* from being read as a policy about vetting *people*.
 var (
-	// escalationVocabPattern recognizes text about collaborator roles and the
-	// permissions those roles carry.
-	escalationVocabPattern = regexp.MustCompile(`(?i)\b(?:maintainer|committer|collaborator|commit access|write access|push access|merge rights|admin(?:istrator)?\s+(?:role|rights|access)|elevated\s+(?:permissions?|privileges?|access)|escalated\s+permissions?|owner\s+role)\b`)
+	escalationAccessVocab = newEscalationVocabulary("access",
+		"write access", "commit access", "push access", "merge right",
+		"elevated permission", "elevated privilege", "elevated access",
+		"escalated permission", "admin access", "admin right",
+		"owner role", "maintainer role", "committer role", "triage right",
+	)
 
-	// reviewVocabPattern recognizes text about a person being reviewed,
-	// approved, nominated, or vetted.
-	reviewVocabPattern = regexp.MustCompile(`(?i)\b(?:review(?:ed)?|approv(?:e|ed|al)|nominat(?:e|ed|ion)|vote(?:d)?|consensus|sponsor(?:ed)?|vett(?:ed|ing))\b`)
+	escalationGrantVocab = newEscalationVocabulary("grant",
+		"grant", "receiv", "promot", "added as", "becom", "onboard",
+		"elevated to",
+	)
 
-	// escalationObligationPattern recognizes language that makes the review a
-	// requirement rather than a description.
-	escalationObligationPattern = regexp.MustCompile(`(?i)\b(?:must|shall|required?|requires|only\s+(?:after|by|through|via)|needs?\s+to\s+be\s+approved)\b`)
+	escalationRoleVocab = newEscalationVocabulary("role",
+		"maintainer", "committer", "collaborator", "approver", "owner",
+	)
 
-	// escalationNegationPattern recognizes language that disclaims or weakens
-	// the requirement within the same section, so contradictory prose is not
-	// credited as a definitive policy (the lesson from the SCA policy prose
-	// checks in vuln_management).
-	escalationNegationPattern = regexp.MustCompile(`(?i)\b(?:not\s+required|no\s+(?:formal\s+)?review|optional|informal(?:ly)?|do(?:es)?\s+not\s+require|without\s+(?:a\s+)?review)\b`)
+	escalationReviewVocab = newEscalationVocabulary("review",
+		"review", "approv", "nominat", "vote", "consensus", "vett",
+		"sponsor", "endors",
+	)
 
-	// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
-	// example snippets cannot satisfy or negate the policy vocabulary.
-	fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
+	escalationObligationVocab = newEscalationVocabulary("obligation",
+		"must", "shall", "requir", "only after", "only by", "only through",
+		"only via", "need to be approv", "may not", "cannot",
+	)
+
+	// An explicit ordering cue is itself a statement of required sequence, so
+	// it also satisfies the obligation gate for declaratively-phrased policies
+	// that carry no modal verb.
+	escalationOrderingVocab = newEscalationVocabulary("ordering",
+		"before", "prior to", "only after", "until", "preced",
+	)
+
+	// Language that disclaims or weakens the requirement, so contradictory
+	// prose is not credited as a definitive policy (the lesson from the SCA
+	// policy prose checks in vuln_management). Evaluated per sentence rather
+	// than per section, so an unrelated "optional" elsewhere cannot suppress a
+	// real policy.
+	escalationNegationVocab = newEscalationVocabulary("negation",
+		"not requir", "no formal review", "no review", "optional",
+		"informal", "does not requir", "do not requir", "without review",
+		"without a review", "at their discretion",
+	)
+
+	// A heading such as "Becoming a Maintainer" establishes that the whole
+	// section concerns conferring a role, which lets sentences inside it omit
+	// the access object without losing the subject.
+	escalationHeadingVocab = newEscalationVocabulary("heading",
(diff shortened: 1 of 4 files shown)


evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
 
 	return gemara.Failed, "No contributor guide documenting requirements for acceptable contributions found in Security Insights data or repository files", gemara.Medium
 }
+
+// Vocabulary for OSPS-GV-04.01. All patterns are word-boundary anchored so
+// unrelated words cannot match as substrings.
+var (
+	// escalationVocabPattern recognizes text about collaborator roles and the
+	// permissions those roles carry.
+	escalationVocabPattern = regexp.MustCompile(`(?i)\b(?:maintainer|committer|collaborator|commit access|write access|push access|merge rights|admin(?:istrator)?\s+(?:role|rights|access)|elevated\s+(?:permissions?|privileges?|access)|escalated\s+permissions?|owner\s+role)\b`)
```
eddie-knight: 🤖 **Minor** — every alternative is followed by `\b`, so plural role nouns never match (`maintainers`, `committers`, `collaborators` all return false), which means a section phrased entirely in plurals such as `New maintainers must be approved by a majority of existing maintainers` produces `mention=false, policy=false` and contributes no signal at all.

```suggestion
	escalationVocabPattern = regexp.MustCompile(`(?i)\b(?:maintainers?|committers?|collaborators?|commit access|write access|push access|merge rights|admin(?:istrator)?\s+(?:roles?|rights|access)|elevated\s+(?:permissions?|privileges?|access)|escalated\s+permissions?|owner\s+roles?)\b`)
```

eddie-knight (commented):
🤖 Nice to see GV-04.01 come off the `NotImplemented` list. I ported `escalationPolicySections` and `classifyEscalationSection` into a standalone harness and ran real documents through them rather than eyeballing the regexes, so the inline notes below cite actual output. Two majors, four minors, one nit.

One thing I checked and want to flag as *clear*: I expected a stock `CODE_OF_CONDUCT.md` to trip this check, but it does not — Contributor Covenant v2.1 run verbatim through the section splitter yields no passing section, since it says "community leaders" throughout and never uses the role nouns in the vocabulary.

evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
...
+
+	// reviewVocabPattern recognizes text about a person being reviewed,
+	// approved, nominated, or vetted.
+	reviewVocabPattern = regexp.MustCompile(`(?i)\b(?:review(?:ed)?|approv(?:e|ed|al)|nominat(?:e|ed|ion)|vote(?:d)?|consensus|sponsor(?:ed)?|vett(?:ed|ing))\b`)
```
eddie-knight: 🤖 **Minor** — same trailing-`\b` issue here: `reviews`, `reviewing`, `approves`, `approving`, and `approvals` all fail to match, so the review gate misses several of the most natural phrasings.

```suggestion
	reviewVocabPattern = regexp.MustCompile(`(?i)\b(?:review(?:ed|s|ing)?|approv(?:e|es|ed|al|als|ing)|nominat(?:e|es|ed|ion|ions)|vote(?:d|s)?|consensus|sponsor(?:ed|s)?|vett(?:ed|ing))\b`)
```

evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
...
+	}
+	flush()
+	return sections
+}
+
+// classifyEscalationSection reports whether a section mentions collaborator
+// escalation together with review (mention), and whether it additionally states
+// the review as an uncontradicted requirement (policy).
+func classifyEscalationSection(section string) (mention bool, policy bool) {
+	if !escalationVocabPattern.MatchString(section) || !reviewVocabPattern.MatchString(section) {
```
eddie-knight: 🤖 **Major** — the two vocabularies are tested independently across the section, so nothing requires the review verb to actually govern the escalation noun, and a stock CONTRIBUTING.md line about reviewing *changes* earns a Pass on a control about vetting *people*: `Every pull request requires approval from a maintainer listed in CODEOWNERS.` returns `mention=true, policy=true`.

The `splitSections` fixture shows you already handle the cross-section version of this, and it does work — my example defeats it only because the escalation noun and the review verb sit in the same sentence. Would it be worth requiring proximity (same sentence or clause), or requiring the escalation gate to match an access-granting phrase like `write access` / `commit access` / `merge rights` rather than accepting a bare role noun? Both are design calls I'd rather leave to you than prescribe.

evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
...
+	if payload.RestData != nil && payload.Insights.Repository != nil &&
+		payload.Insights.Repository.Documentation != nil &&
+		payload.Insights.Repository.Documentation.Governance != nil {
+		return gemara.NeedsReview, "Governance documentation is declared in Security Insights; confirm it requires review of code collaborators before escalated permissions are granted", gemara.Medium
+	}
+	if file := governanceFilePresent(files); file != "" {
+		return gemara.NeedsReview, "A governance document (" + file + ") is present but no escalation-review policy language was recognized; manual review required", gemara.Low
+	}
+	if docsErr != nil {
+		return gemara.NeedsReview, "Repository documentation could not be fully read; manually review whether a collaborator escalation-review policy is documented", gemara.Low
```
eddie-knight: 🤖 **Major** — `payload.InsightsError` is never consulted, so when `security-insights.yml` is present but fails to parse (`data/rest-data.go:399`) `ensureInsightsInitialized` leaves `Insights.Repository.Documentation` nil, the governance branch above cannot fire, and the step falls through to `gemara.Failed` citing "Security Insights data" that was never successfully read — contradicting this function's own doc comment that unreadable documentation is never reported as a violation.

The sibling checks guard exactly this at `evaluation_plans/osps/vuln_management/steps.go:1622` and `:1696`.

```suggestion
	if docsErr != nil || (payload.RestData != nil && payload.InsightsError) {
		return gemara.NeedsReview, "Repository documentation or Security Insights could not be completely inspected; manually review whether a collaborator escalation-review policy is documented", gemara.Low
```

evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
...
+// escalation together with review (mention), and whether it additionally states
+// the review as an uncontradicted requirement (policy).
+func classifyEscalationSection(section string) (mention bool, policy bool) {
+	if !escalationVocabPattern.MatchString(section) || !reviewVocabPattern.MatchString(section) {
+		return false, false
+	}
+	if !escalationObligationPattern.MatchString(section) {
+		return true, false
+	}
+	if escalationNegationPattern.MatchString(section) {
```
eddie-knight: 🤖 **Minor** — the negation pattern is evaluated over the whole section rather than the matching clause, so an unrelated sentence like `Adding yourself to the AUTHORS file is optional.` downgrades a genuine, correctly-stated policy in the same section to `NeedsReview`.

evaluation_plans/osps/governance/steps_test.go
```diff
@@ -355,3 +355,143 @@ func TestHasRolesAndResponsibilities(t *testing.T) {
...
+			Name:    github.Ptr(fileName),
+			Path:    github.Ptr(fileName),
+			Content: github.Ptr(content),
+		}},
+		map[string][]*github.RepositoryContent{},
+	)
+}
+
+func TestHasEscalatedPermissionsReviewPolicy(t *testing.T) {
+	const clearPolicy = "# Becoming a maintainer\n\nNew maintainers must be approved by two existing maintainers before receiving write access.\n"
```
eddie-knight, line 375: 🤖 **Minor** — this fixture (and the table cases below it) pair plural role nouns with `write access`, and it is `write access` that actually carries the escalation match, so the suite stays green while the plural gap flagged above remains invisible; a fixture using only plural role nouns and no access phrase would fail today.

evaluation_plans/osps/governance/steps.go
```diff
@@ -146,3 +147,137 @@ func HasContributionReviewPolicy(payload data.Payload) (result gemara.Result, me
...
+	escalationObligationPattern = regexp.MustCompile(`(?i)\b(?:must|shall|required?|requires|only\s+(?:after|by|through|via)|needs?\s+to\s+be\s+approved)\b`)
+
+	// escalationNegationPattern recognizes language that disclaims or weakens
+	// the requirement within the same section, so contradictory prose is not
+	// credited as a definitive policy (the lesson from the VM-05 prose checks).
+	escalationNegationPattern = regexp.MustCompile(`(?i)\b(?:not\s+required|no\s+(?:formal\s+)?review|optional|informal(?:ly)?|do(?:es)?\s+not\s+require|without\s+(?:a\s+)?review)\b`)
+
+	// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
+	// example snippets cannot satisfy or negate the policy vocabulary.
+	fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
```
eddie-knight: 🤖 **Nit** — the sibling `documentationSections` (`evaluation_plans/osps/vuln_management/steps.go`) tracks fence state as it scans instead of regex-stripping, which also survives an unterminated code fence; not wrong as written, but a third consumer of this pattern seems like a reasonable trigger to extract the shared helper.

evaluation_plans/osps/governance/steps.go
eddie-knight: 🤖 **Alternate, preferred suggestion:** replace the four regexes with a tokenized vocabulary. I don't think the major above is separately patchable — four independent `MatchString` calls over a section discard which words relate to which, and whether something is *granted* or merely *named*. The plural gap has the same root cause: `\b`-suffixed alternatives are what you reach for when the vocabulary lives inside a pattern instead of being data.

Four changes:

1. **Vocabulary becomes data.** `newEscalationVocabulary("access", "write access", …)` — extend by adding a string, not by editing an alternation, and each set is testable on its own.
2. **Entries are stems, matched by prefix.** `approv` covers approve/approved/approval/approving; `maintainer` covers the plural. The whole inflection bug class disappears rather than being patched. Entries under four characters (`to`, `no`) must match exactly, so they can't match inside unrelated words.
3. **`find` returns a position, not a bool.** This is the one that matters: the control requires review *prior to* granting, and an ordering relation cannot be expressed with a bool.
4. **Role nouns are separated from access objects.** A bare `maintainer` is not escalation; `write access`, or a granting verb applied to a role, is. This is what stops `Every pull request requires approval from a maintainer` — prose about reviewing a change — from being read as a policy about vetting people.

Scope also moves from section to sentence, which is what ties the review verb to the escalation it governs, and makes negation local so an unrelated "…is optional" no longer suppresses a real policy.

**Verified, not asserted.** `classifyEscalationSection` and `escalationPolicySections` keep their exact signatures, so **your test file passes unmodified** — `clearPolicy`, `negatedPolicy`, `mentionOnly`, `splitSections`, `fencedOnly` and all five table rows. Full repository suite green, `go vet` and `gofmt` clean. It additionally rejects the two false positives and recovers the three false negatives noted above.

<details>
<summary><b>Patch</b> — <code>evaluation_plans/osps/governance/steps.go</code> (+192 / −27)</summary>

````diff
@@ -148,29 +148,135 @@
 	return gemara.Failed, "No contributor guide documenting requirements for acceptable contributions found in Security Insights data or repository files", gemara.Medium
 }
 
-// Vocabulary for OSPS-GV-04.01. All patterns are word-boundary anchored so
-// unrelated words cannot match as substrings.
+// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
+// example snippets cannot satisfy or negate the policy vocabulary.
+var fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
+
+// escalationTokens lowercases prose and strips punctuation. No stemming is
+// attempted: vocabulary entries are written as stems and compared by prefix,
+// which handles inflection by construction rather than by guessing suffixes.
+func escalationTokens(text string) []string {
+	fields := strings.Fields(text)
+	tokens := make([]string, 0, len(fields))
+	for _, field := range fields {
+		token := strings.ToLower(strings.Trim(field, ".,;:!?()[]{}\"'`"))
+		if token != "" {
+			tokens = append(tokens, token)
+		}
+	}
+	return tokens
+}
+
+// escalationTokenMatches compares one prose token against one vocabulary token.
+// A vocabulary entry of four or more characters is treated as a stem, so
+// "approv" covers approve/approved/approval/approves/approving and "maintainer"
+// covers the plural, without enumerating forms. Short entries such as "to" must
+// match exactly, so they cannot match inside unrelated words.
+func escalationTokenMatches(token, vocabToken string) bool {
+	if len(vocabToken) >= 4 {
+		return strings.HasPrefix(token, vocabToken)
+	}
+	return token == vocabToken
+}
+
+// escalationVocabulary is a set of phrases, each one or more consecutive
+// normalized tokens. Extending a vocabulary means adding a string.
+type escalationVocabulary struct {
+	name    string
+	phrases [][]string
+}
+
+func newEscalationVocabulary(name string, entries ...string) escalationVocabulary {
+	vocabulary := escalationVocabulary{name: name}
+	for _, entry := range entries {
+		vocabulary.phrases = append(vocabulary.phrases, escalationTokens(entry))
+	}
+	return vocabulary
+}
+
+// find returns the index of the earliest matching phrase, or -1. It returns a
+// position rather than a bool because the control requires review *prior to*
+// escalation, and an ordering relation cannot be expressed with a bool.
+func (v escalationVocabulary) find(tokens []string) int {
+	earliest := -1
+	for _, phrase := range v.phrases {
+		if len(phrase) == 0 {
+			continue
+		}
+		for i := 0; i+len(phrase) <= len(tokens); i++ {
+			matched := true
+			for j, want := range phrase {
+				if !escalationTokenMatches(tokens[i+j], want) {
+					matched = false
+					break
+				}
+			}
+			if matched && (earliest == -1 || i < earliest) {
+				earliest = i
+			}
+		}
+	}
+	return earliest
+}
+
+func (v escalationVocabulary) has(tokens []string) bool { return v.find(tokens) >= 0 }
+
+// Vocabulary for OSPS-GV-04.01. Note the distinction a single pattern cannot
+// draw: a role noun ("maintainer") is not itself evidence of escalation, while
+// an access object ("write access") or a granting verb applied to a role is.
+// That separation is what keeps prose about reviewing *changes* from being read
+// as a policy about vetting *people*.
 var (
-	// escalationVocabPattern recognizes text about collaborator roles and the
-	// permissions those roles carry.
-	escalationVocabPattern = regexp.MustCompile(`(?i)\b(?:maintainer|committer|collaborator|commit access|write access|push access|merge rights|admin(?:istrator)?\s+(?:role|rights|access)|elevated\s+(?:permissions?|privileges?|access)|escalated\s+permissions?|owner\s+role)\b`)
+	escalationAccessVocab = newEscalationVocabulary("access",
+		"write access", "commit access", "push access", "merge right",
+		"elevated permission", "elevated privilege", "elevated access",
+		"escalated permission", "admin access", "admin right",
+		"owner role", "maintainer role", "committer role", "triage right",
+	)
 
-	// reviewVocabPattern recognizes text about a person being reviewed,
-	// approved, nominated, or vetted.
-	reviewVocabPattern = regexp.MustCompile(`(?i)\b(?:review(?:ed)?|approv(?:e|ed|al)|nominat(?:e|ed|ion)|vote(?:d)?|consensus|sponsor(?:ed)?|vett(?:ed|ing))\b`)
+	escalationGrantVocab = newEscalationVocabulary("grant",
+		"grant", "receiv", "promot", "added as", "becom", "onboard",
+		"elevated to",
+	)
 
-	// escalationObligationPattern recognizes language that makes the review a
-	// requirement rather than a description.
-	escalationObligationPattern = regexp.MustCompile(`(?i)\b(?:must|shall|required?|requires|only\s+(?:after|by|through|via)|needs?\s+to\s+be\s+approved)\b`)
+	escalationRoleVocab = newEscalationVocabulary("role",
+		"maintainer", "committer", "collaborator", "approver", "owner",
+	)
 
-	// escalationNegationPattern recognizes language that disclaims or weakens
-	// the requirement within the same section, so contradictory prose is not
-	// credited as a definitive policy (the lesson from the VM-05 prose checks).
-	escalationNegationPattern = regexp.MustCompile(`(?i)\b(?:not\s+required|no\s+(?:formal\s+)?review|optional|informal(?:ly)?|do(?:es)?\s+not\s+require|without\s+(?:a\s+)?review)\b`)
+	escalationReviewVocab = newEscalationVocabulary("review",
+		"review", "approv", "nominat", "vote", "consensus", "vett",
+		"sponsor", "endors",
+	)
 
-	// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
-	// example snippets cannot satisfy or negate the policy vocabulary.
-	fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
+	escalationObligationVocab = newEscalationVocabulary("obligation",
+		"must", "shall", "requir", "only after", "only by", "only through",
+		"only via", "need to be approv", "may not", "cannot",
+	)
+
+	// An explicit ordering cue is itself a statement of required sequence, so
+	// it also satisfies the obligation gate for declaratively-phrased policies
+	// that carry no modal verb.
+	escalationOrderingVocab = newEscalationVocabulary("ordering",
+		"before", "prior to", "only after", "until", "preced",
+	)
+
+	// Language that disclaims or weakens the requirement, so contradictory
+	// prose is not credited as a definitive policy (the lesson from the VM-05
+	// prose checks). Evaluated per sentence rather than per section, so an
+	// unrelated "optional" elsewhere cannot suppress a real policy.
+	escalationNegationVocab = newEscalationVocabulary("negation",
+		"not requir", "no formal review", "no review", "optional",
+		"informal", "does not requir", "do not requir", "without review",
+		"without a review", "at their discretion",
+	)
+
+	// A heading such as "Becoming a Maintainer" establishes that the whole
+	// section concerns conferring a role, which lets sentences inside it omit
+	// the access object without losing the subject.
+	escalationHeadingVocab = newEscalationVocabulary("heading",
+		"becom", "adding", "onboard", "promotion", "nominat", "joining",
+		"new maintainer", "new committer",
+	)
 )
 
 // escalationPolicySections splits a documentation file into markdown-heading
@@ -196,20 +302,79 @@
 	return sections
 }
 
+// escalationSentences splits a section into sentences. Judging co-occurrence
+// per sentence rather than per section is what ties the review verb to the
+// escalation it governs.
+func escalationSentences(section string) []string {
+	var sentences []string
+	var current strings.Builder
+	flush := func() {
+		if sentence := strings.TrimSpace(current.String()); sentence != "" {
+			sentences = append(sentences, sentence)
+		}
+		current.Reset()
+	}
+	for _, r := range section {
+		current.WriteRune(r)
+		if r == '.' || r == ';' || r == '!' || r == '\n' {
+			flush()
+		}
+	}
+	flush()
+	return sentences
+}
+
 // classifyEscalationSection reports whether a section mentions collaborator
 // escalation together with review (mention), and whether it additionally states
 // the review as an uncontradicted requirement (policy).
+//
+// A section carries its own heading as its first line, so heading context is
+// available without changing the caller.
 func classifyEscalationSection(section string) (mention bool, policy bool) {
-	if !escalationVocabPattern.MatchString(section) || !reviewVocabPattern.MatchString(section) {
-		return false, false
+	heading, _, _ := strings.Cut(section, "\n")
+	headingTokens := escalationTokens(heading)
+	headingConfersRole := escalationHeadingVocab.has(headingTokens) &&
+		escalationRoleVocab.has(headingTokens)
+
+	for _, sentence := range escalationSentences(section) {
+		tokens := escalationTokens(sentence)
+
+		reviewAt := escalationReviewVocab.find(tokens)
+		if reviewAt < 0 {
+			continue
+		}
+
+		// The escalation subject must be an access object, or a granting verb
+		// applied to a role -- not a bare role noun. This is what stops
+		// "requires approval from a maintainer", which is about reviewing a
+		// change, from counting as evidence about granting access to a person.
+		escalationAt := escalationAccessVocab.find(tokens)
+		if escalationAt < 0 && escalationGrantVocab.has(tokens) && escalationRoleVocab.has(tokens) {
+			escalationAt = escalationGrantVocab.find(tokens)
+		}
+		if escalationAt < 0 && !headingConfersRole {
+			continue
+		}
+
+		mention = true
+
+		explicitOrder := escalationOrderingVocab.has(tokens)
+		if !escalationObligationVocab.has(tokens) && !explicitOrder {
+			continue
+		}
+		if escalationNegationVocab.has(tokens) {
+			continue
+		}
+		// Reviewed prior to granting: an explicit cue is strongest, otherwise
+		// the review term appearing ahead of the escalation term is weaker
+		// positional evidence of the same ordering.
+		if !explicitOrder && escalationAt >= 0 && reviewAt > escalationAt {
+			continue
+		}
+		policy = true
 	}
-	if !escalationObligationPattern.MatchString(section) {
-		return true, false
-	}
-	if escalationNegationPattern.MatchString(section) {
-		return true, false
-	}
-	return true, true
+
+	return mention, policy
 }
 
 // HasEscalatedPermissionsReviewPolicy implements OSPS-GV-04.01: while active,
````

</details>

Remaining costs, so they aren't a surprise: prefix matching means `maintainer` also matches `maintainership`, sentence splitting on `.` will mishandle `e.g.`, and this is still heuristic prose matching — `gemara.Medium` remains the honest confidence.

**Comment by jmeridth** [comment]
🤖 7234d5a rewords this PR's new doc comments per #448 category 1: behavior descriptions instead of catalog IDs, and sibling-check references now use the function name (`HasContributionReviewPolicy`) rather than the bare requirement ID.

File: evaluation_plans/osps/governance/steps.go (hunk 1 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732. Adopted the tokenized vocabulary from your follow-up, so stems compared by prefix cover the plural forms by construction rather than by adding s? alternations.

File: evaluation_plans/osps/governance/steps.go (hunk 2 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732. Same root cause as the role-noun gap; the review vocabulary is now stems (review, approv, nominat) matched by prefix, which covers reviews, reviewing, approves, approving, and approvals.

File: evaluation_plans/osps/governance/steps.go (hunk 3 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Adopted your preferred patch in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732. Good catch on the same-sentence defeat of the section splitting. I verified the claims independently: the existing fixtures pass unmodified against the new classifier, and a new fixture locks the rejection of the CODEOWNERS example (review of changes is not a policy about vetting people). One deviation: the comment above the vocabulary block describes the check's behavior instead of naming the requirement ID, per the convention in #448.

File: evaluation_plans/osps/governance/steps.go (hunk 4 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732 with your suggested guard, plus a regression case (unparseable security insights needs review instead of failing) so the doc comment's promise is now enforced by a test.

File: evaluation_plans/osps/governance/steps.go (hunk 5 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732. Negation is evaluated per sentence in the tokenized classifier, so an unrelated optional elsewhere in the section no longer suppresses a genuine policy.

File: evaluation_plans/osps/governance/steps_test.go (hunk 6 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps_test.go:375]: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/8233cd188caf3238855a0993f30c12323c618732. Added a plural-only fixture with no access phrase (plural role nouns without an access phrase still pass), which fails against the old regexes and passes with the tokenizer.

File: evaluation_plans/osps/governance/steps.go (hunk 7 above)
Comment (jmeridth) [reply evaluation_plans/osps/governance/steps.go]: @eddie-knight Agreed on the trigger, but leaving this one out of this PR to keep it reviewable. The shared fence-aware section helper fits the cleanup pass tracked in #448, where the other shared-helper work already lives; I have added it to that issue's checklist.

**Review by eddie-knight — DISMISSED** [review]
🤖 Jason asked for a final follow-up pass. Approving — all seven items from the earlier review are addressed, and the one note below is the only thing I could find.

Verified against `8233cd1` on the extracted tree rather than by reading: `go build ./...`, `go vet ./...` and `gofmt -l` clean, **full repository suite green**, and the five adversarial cases I wrote *before* these changes pass unmodified against the new classifier. PR scope is still the same four files.

| Finding | Status |
| --- | --- |
| Plural role nouns never match | Resolved — stems matched by prefix |
| Plural/inflected review terms never match | Resolved — same mechanism |
| Code-review prose credited as escalation policy | Resolved — role nouns separated from access objects |
| `InsightsError` not consulted | Resolved — guard matches the sibling's shape |
| Negation evaluated over the whole section | Resolved — now per sentence |
| Fixtures masked the plural gap | Resolved — `plural role nouns without an access phrase still pass` |
| Fence stripping / shared helper | Extraction deferred to #448 — agreed; see below |

**One residual, non-blocking.** Deferring the shared fence-aware helper to #448 makes sense and I'm not reopening it. But that nit had two halves, and the correctness half is independent of extraction: `fencedCodeBlockPattern` needs a *closing* fence to strip anything, so an unterminated fence leaves its contents to be read as prose. On this tree:

````text
# Setup

```
New maintainers must be approved before receiving write access.
````

→ policy credited: **true** (fence stripped: false)

Tracking fence state in the loop that already walks lines for headings fixes it by construction — an unterminated fence stays open to EOF, so the remainder is skipped, which is the safe direction for a malformed document. It also picks up `~~~` fences, which the pattern never matched, and removes the file's only `regexp` use. 18 lines added, 10 removed, no extraction involved, so it shouldn't cost you the reviewability you were protecting — and #448 can still subsume it later.

<details>
<summary><b>Patch</b> — <code>evaluation_plans/osps/governance/steps.go</code> (+18 / −10)</summary>

````diff
@@ -1,7 +1,6 @@
 package governance
 
 import (
-	"regexp"
 	"strings"
 
 	"github.com/gemaraproj/go-gemara"
@@ -148,10 +147,6 @@
 	return gemara.Failed, "No contributor guide documenting requirements for acceptable contributions found in Security Insights data or repository files", gemara.Medium
 }
 
-// fencedCodeBlockPattern strips fenced code blocks before prose analysis so
-// example snippets cannot satisfy or negate the policy vocabulary.
-var fencedCodeBlockPattern = regexp.MustCompile("(?s)```.*?```")
-
 // escalationTokens lowercases prose and strips punctuation. No stemming is
 // attempted: vocabulary entries are written as stems and compared by prefix,
 // which handles inflection by construction rather than by guessing suffixes.
@@ -284,17 +279,30 @@
 // sections with fenced code blocks removed, so vocabulary co-occurrence is
 // judged within one topical section rather than across an entire file.
 func escalationPolicySections(content string) []string {
-	cleaned := fencedCodeBlockPattern.ReplaceAllString(content, "")
-	var sections []string
-	var current []string
+	var (
+		sections []string
+		current  []string
+		inFence  bool
+	)
 	flush := func() {
 		if len(current) > 0 {
 			sections = append(sections, strings.Join(current, "\n"))
 			current = nil
 		}
 	}
-	for _, line := range strings.Split(cleaned, "\n") {
-		if strings.HasPrefix(strings.TrimSpace(line), "#") {
+	for _, line := range strings.Split(content, "\n") {
+		trimmed := strings.TrimSpace(line)
+		// Tracking fence state as we scan, rather than matching balanced pairs,
+		// means an unterminated fence swallows the rest of the file instead of
+		// leaving its contents to be read as prose. It also covers ~~~ fences.
+		if strings.HasPrefix(trimmed, "```") || strings.HasPrefix(trimmed, "~~~") {
+			inFence = !inFence
+			continue
+		}
+		if inFence {
+			continue
+		}
+		if strings.HasPrefix(trimmed, "#") {
 			flush()
 		}
 		current = append(current, line)
````

</details>

Six fence cases pass with it (closed, unterminated, info-string, `~~~`, prose-after-a-closed-fence still read, plain prose unaffected) and your existing `fencedOnly` fixture is unchanged. Happy to hand over that test file too.

One thing I deliberately left out: the sibling also skips blockquote lines, and I didn't copy that — a doc putting its policy in a blockquote would silently stop being credited, so it seemed like your call rather than something to fold into a fence fix.

jmeridth:
🤖 @eddie-knight

> **One residual, non-blocking.** ... `fencedCodeBlockPattern` needs a *closing* fence to strip anything, so an unterminated fence leaves its contents to be read as prose.

Both halves addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/163e346. Took your fence patch as written: stateful tracking, tilde fences covered, unterminated fences swallow to EOF, and the file's last regexp use is gone.

On the blockquote question you left open: went with skipping them, matching the vuln_management scanner. The deciding case is a governance doc quoting the baseline requirement it intends to satisfy. That text contains a review verb, an access object, an obligation, and an ordering cue, so it credits policy=true verbatim. The miss direction degrades softly (a callout-only policy in GOVERNANCE.md falls to the governance-file NeedsReview branch), while the hit direction prevented an unearned Pass. Four fixtures lock it in: unterminated fence, tilde fence, blockquoted requirement text, and a callout-only policy documenting the intentional NeedsReview outcome. Consistency also keeps the shared-helper extraction in #448 possible, since the two scanners now agree on both fence and blockquote semantics.

eddie-knight (approved):

PR #447, opened 2026-08-14
fix(OSPS-QA-07.01): stop false Failed from invisible branch protection

### LINKED ISSUE #440: False positive: OSPS-QA-07
It seems as if the check for required reviewers is misbehaving, potentially due to token permission requirements

jmeridth:
Fixes #440

## What/Why

`RequiresNonAuthorApproval` (OSPS-QA-07.01) read only admin-gated classic branch protection GraphQL fields, producing false `Failed` verdicts in two independent ways: any non-admin scan of a classically protected repository (the fields return zero values indistinguishable from no protection), and any scan of a ruleset-protected repository regardless of token (classic `branchProtectionRule` is null when protection comes from rulesets). Live probes reproduced both: docker/cli through a READ token, and this repository itself through a MAINTAIN token, both fail today while their public ruleset endpoints require one approval.

The step now aggregates the publicly readable ruleset `pull_request` rules across every applying rule (max approval count, any-rule booleans, so the verdict is not rule-order dependent), falls back to classic protection when observed, returns `NeedsReview` rather than `Failed` when protection is unobservable (the `access_control` convention for exactly this), and treats the stale-approval gap (approvals required but neither last-push approval nor stale-review dismissal enabled) as `NeedsReview` instead of the previous hard `Failed`, since the catalog text requires only one non-author approval and GitHub already forbids self-approval.

## Proof it works

- The two live false-positive shapes are now table-test fixtures; the previous implementation had zero test coverage, which is how this shipped unnoticed.
- 8-case step table asserting result, message, and confidence for every branch of the new matrix, plus a data-layer test for multi-rule aggregation including nil entries.
- `go test ./...` across all packages and `go test -race` on the touched packages pass.

## Risk + AI role

Low: one step rewritten plus one additive `RepositoryMetadata` accessor; no other step's behavior changes. AI-generated end to end with maintainer-approved diagnosis and fix direction.

## Review focus

- The verdict matrix, especially `NeedsReview` rather than `Failed` for non-admin scans with no observed rules.
- Treating the stale-approval gap as `NeedsReview` rather than a hard `Failed`.


jmeridth:
🤖 ecab5d1 rewords this PR's new doc comment to describe behavior instead of the catalog ID, per #448 category 1. The ID linkage stays in the dispatch map entry.

eddie-knight (commented):
🤖 **Multi-model review** — three independent reviewers (haiku, sonnet, opus) each reviewed the full worktree at 4ac57bd; every claim was re-verified against source by the orchestrator before posting.

**Focus: idiomatic, DRY, functional?**

- **Idiomatic** — yes, unanimous. The verdict ladder matches the repo's access_control convention (`access_control/steps.go:78-95`: public rulesets first, `!ViewerCanAdminister()` → NeedsReview, Failed only when both sources are observable), the accessor mirrors `RequiredStatusCheckContexts`, and each test file follows its own file's style (`t.Errorf` in the quality steps test, testify in the data test).
- **DRY** — within the PR, yes; no duplicated logic. One pre-existing overlap is worth a follow-up issue (not this PR): `DefaultBranchRequiresPRReviews` (data/repository_metadata.go:78-83) still reads only `PullRequest[0]` — the same rule-order dependency this PR eliminates — and is live via `BranchProtectionRestrictsPushes` (`access_control/steps.go:62`). Two reviewers raised it independently; confirmed by reading the method body. It could collapse to a one-liner over the new accessor.
- **Functional** — yes. Two reviewers independently ran `go build ./...`, `go vet ./...`, `go test ./...`, and `go test -race` on the touched packages (all pass, `gofmt -l` clean), and traced all 8 table cases through the ladder. Field names match go-github v74 (`rules.go:416-425`), and both judgment calls the PR flags for review are grounded in the catalog text (`OSPS_Baseline_2026_02.yaml:2422-2426` requires only one non-author approval, with no re-approval clause) — NeedsReview is the right verdict for both the unobservable case and the stale-approval gap.

**Findings**: nothing blocking; merge-ready per all three reviewers. Seven inline comments: the two Minor items worth doing before merge are the `classicCount` fold that disagrees with its own gate (steps.go:193) and the two untested matrix branches; the rest are the comment-shortening suggestions requested plus small wording touches.

**Checked and NOT problems** (recorded so they don't get re-raised):

- Max/any-boolean aggregation in `DefaultBranchPullRequestReviewRules` is order-independent and nil-entry-safe — one reviewer filed this as a defect and refuted it within the same report; confirmed against the multi-rule test.
- The Failed message when classic protection exists with a required count of 0 is factually accurate ("no approving review required"), so a claim that it lost the old "requires 0 approving reviews" diagnostic was checked and dropped.
- Interface widening is compile-safe: every existing `RepositoryMetadata` test fake embeds the interface. Production nil-safety is guaranteed upstream (`data/payload.go` builds both payload halves in an errgroup that aborts on error); the new guards are defense matching `vuln_management`.
- No sibling step's verdict shifts: OSPS-QA-07.01 runs this step alone, and the untouched access_control steps read accessors this PR does not modify.

evaluation_plans/osps/quality/steps.go
```diff
@@ -150,23 +150,65 @@ func NoUnreviewableBinariesInRepo(payload data.Payload) (result gemara.Result, m
...
+		classicRequires = protection.RequiresApprovingReviews
+		classicCount = payload.Repository.DefaultBranchRef.RefUpdateRule.RequiredApprovingReviewCount
+		classicLastPush = protection.RequireLastPushApproval
 	}
 
-	if !protection.RequireLastPushApproval {
-		return gemara.Failed, "Branch protection does not require re-approval after new commits", confidence
+	if ruleset.RequiredApprovals >= 1 || (classicRequires && classicCount >= 1) {
+		approvals := ruleset.RequiredApprovals
+		if classicCount > approvals {
```
eddie-knight: 🤖 **Minor** — The gate at line 191 trusts `classicCount` only when `classicRequires` is true, but this fold uses it unconditionally: a repo entering via the ruleset arm with `classicRequires=false` and a stale non-zero `RefUpdateRule` count reports a number the gate itself refused to trust.

```suggestion
		if classicRequires && classicCount > approvals {
```

data/repository_metadata.go
```diff
@@ -19,9 +19,24 @@ type RepositoryMetadata interface {
...
 
+// PullRequestReviewRules summarizes the pull_request rules the repository's
+// rulesets enforce on the default branch. Observed distinguishes a completed
+// ruleset lookup from one that never happened, and RuleCount distinguishes
+// "observed with no pull_request rule" from unobserved. When several rulesets
+// apply, GitHub enforces all of them, so RequiredApprovals is the maximum
+// declared count and the booleans are true when any applying rule sets them.
+type PullRequestReviewRules struct {
+	Observed                bool
+	RuleCount               int
```
eddie-knight: 🤖 **Minor** — `RuleCount` is never read outside tests (grep hits only this struct, the increment, and test assertions), and the doc's justification for it — distinguishing "observed with no pull_request rule" from unobserved — is what `Observed` already does. Deleting the field, the `rules.RuleCount++`, and the test assertions on it removes the dead weight; the assertions can retarget `RequiredApprovals`.

evaluation_plans/osps/quality/steps_test.go
```diff
@@ -1127,3 +1127,151 @@ func TestIsSignatureOrChecksumAsset(t *testing.T) {
...
+			wantConfidence: gemara.Low,
+		},
+		{
+			name:           "zero-value payload does not panic and needs review",
+			payload:        data.Payload{},
+			wantResult:     gemara.NeedsReview,
+			wantMsgPart:    "only visible to admin tokens",
+			wantConfidence: gemara.Low,
+		},
+	}
```
eddie-knight, line 1313: 🤖 **Minor** — Two branches of the matrix have no case: (a) the classic-only stale gap — `classic(true, 1, false)` with no ruleset approvals should hit the NeedsReview return at steps.go:200 via the classic arm, and (b) the cross-source max with both sources non-zero — the only case exercising the fold uses ruleset 0 / classic 2, so `ruleset: 1, classic: 3` asserting "requires 3" would pin the aggregation. Two more table rows cover both.

evaluation_plans/osps/quality/steps.go
```diff
@@ -150,23 +150,65 @@ func NoUnreviewableBinariesInRepo(payload data.Payload) (result gemara.Result, m
...
+// visible to admin tokens: for anyone else the GraphQL objects come back as
+// zero values indistinguishable from no protection, so an unobserved absence
+// is reported as NeedsReview rather than Failed (see issue #440, where both a
+// READ-token scan of docker/cli and a MAINTAIN-token scan of this repository
+// produced false Failed verdicts).
+//
+// When a review requirement exists but neither last-push approval nor
+// stale-review dismissal is enabled, commits pushed after an approval can merge
+// unreviewed. The catalog text only requires one non-author approval, so that
+// gap surfaces as NeedsReview rather than the previous hard Failed.
```
eddie-knight: 🤖 **Minor** — This doc comment is 20 lines on a 39-line function, 3-4x any sibling in the file, and the middle paragraph restates the `PullRequestReviewRules` type doc. It reads well but the same information survives at half the length:

```suggestion
// RequiresNonAuthorApproval checks that changes require at least one
// non-author approving review before merging to the default branch. GitHub
// forbids self-approval, so a required count >= 1 satisfies the catalog.
//
// Rulesets are publicly readable and aggregated across every applying rule.
// Classic branch protection is admin-only and reads as zero values otherwise,
// so an unobserved absence is NeedsReview rather than Failed (#440). A
// requirement without last-push approval or stale-review dismissal lets
// commits pushed after an approval merge unreviewed; the catalog does not
// cover that gap, so it surfaces as NeedsReview.
```

data/repository_metadata.go
```diff
@@ -140,6 +155,29 @@ func (r *GitHubRepositoryMetadata) RulesetsObserved() bool {
 	return r.defaultBranchRules != nil
 }
 
+// DefaultBranchPullRequestReviewRules aggregates every pull_request rule the
+// default branch's rulesets enforce. Aggregation matters: multiple rulesets can
+// apply simultaneously and GitHub enforces the union, so reading only the first
+// rule would make the result depend on rule order.
```
eddie-knight: 🤖 **Nit** — This repeats the union/order-dependency rationale already in the type doc above; a pointer is enough.

```suggestion
// DefaultBranchPullRequestReviewRules aggregates every applying pull_request
// rule; see PullRequestReviewRules for the aggregation semantics.
```

evaluation_plans/osps/quality/steps.go
```diff
@@ -150,23 +150,65 @@ func NoUnreviewableBinariesInRepo(payload data.Payload) (result gemara.Result, m
...
-	if !protection.RequireLastPushApproval {
-		return gemara.Failed, "Branch protection does not require re-approval after new commits", confidence
+	if ruleset.RequiredApprovals >= 1 || (classicRequires && classicCount >= 1) {
+		approvals := ruleset.RequiredApprovals
+		if classicCount > approvals {
+			approvals = classicCount
+		}
+		staleGapClosed := ruleset.RequireLastPushApproval || ruleset.DismissStaleReviews || classicLastPush
+		if staleGapClosed {
+			return gemara.Passed, fmt.Sprintf("The default branch requires %d non-author approving review(s), and approvals cannot carry commits pushed after the review to merge", approvals), gemara.High
```
eddie-knight: 🤖 **Nit** — "approvals cannot carry commits pushed after the review to merge" is hard to parse in a user-facing message; line 200 already has the clearer phrasing.

```suggestion
			return gemara.Passed, fmt.Sprintf("The default branch requires %d non-author approving review(s), and commits pushed after an approval cannot merge unreviewed", approvals), gemara.High
```

evaluation_plans/osps/quality/steps.go
```diff
@@ -150,23 +150,65 @@ func NoUnreviewableBinariesInRepo(payload data.Payload) (result gemara.Result, m
...
+		}
+		return gemara.NeedsReview, fmt.Sprintf("The default branch requires %d non-author approving review(s), but neither last-push approval nor stale-review dismissal is enabled, so commits pushed after an approval can merge unreviewed; confirm the review process covers this gap", approvals), gemara.Medium
 	}
 
-	return gemara.Passed, fmt.Sprintf("Branch protection requires %d approving reviews and re-approval after new commits", reviewCount), confidence
+	// No review requirement was observed anywhere. Classic branch protection is
+	// invisible to non-admin tokens, so absence cannot be asserted without admin.
+	if !adminObservable {
```
eddie-knight: 🤖 **Nit** — Because `!adminObservable` is checked first, the more specific "rulesets could not be observed" message below is unreachable for non-admin tokens — a non-admin scan whose ruleset fetch also failed reports only the admin-visibility reason. Checking `!ruleset.Observed` first (or mentioning both conditions here) keeps the message accurate in that state.

eddie-knight (dismissed):

evaluation_plans/osps/quality/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd with your suggestion, plus a red-first matrix row (stale classic count without classic requirement is not reported) pinning that a ruleset-arm entry reports the ruleset count.

data/repository_metadata.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd. Field, increment, doc mention, and both test files' assertions are gone; the two Zero assertions retarget RequiredApprovals.

evaluation_plans/osps/quality/steps_test.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd. Both rows added: classic(true, 1, false) with an empty ruleset hits the stale-gap NeedsReview, and ruleset 1 with classic 3 asserts requires 3.

evaluation_plans/osps/quality/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd with your replacement as written.

data/repository_metadata.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd with your replacement as written.

evaluation_plans/osps/quality/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd with your wording.

evaluation_plans/osps/quality/steps.go
jmeridth: @eddie-knight Addressed in https://github.com/ossf/pvtr-github-repo-scanner/commit/83a9c436761bd1405c6686c5ee9e309575622acd by checking the ruleset gap first, since it applies to any token while the admin caveat only explains the classic blind spot. One knock-on made explicit in the tests: the nil-metadata zero-value payload now also reports the ruleset message, since nothing was looked up in that state either.

eddie-knight (approved):
Three non-blocking items surfaced by opus and verified by fable:

1. Doc/policy mismatch (in-diff, steps.go:153-162): the comment says a required count >= 1 "satisfies the catalog," but the code returns NeedsReview for that case when the stale gap is open. The draft includes a drop-in suggestion rewording it.
2. Classic arm can't see stale-review dismissal (steps.go:186): dismissesStaleReviews isn't in the GraphQL query, so a classic-BP repo with dismissal-only lands NeedsReview. Strictly better than the old hard Failed — fast-follow material.
3. Pre-existing, out of diff: DefaultBranchRequiresPRReviews still reads only PullRequest[0] (order-dependent, unguarded nil element) — the exact pattern the PR's new accessor was written to avoid, and it feeds OSPS-AC. One-line delegation fixes it; belongs in a follow-up and intersects the #459 metadata work.

jmeridth:
@eddie-knight I'm going to do a quick-follow PR for 1 and 2 above. 3 can be handled by #459

jmeridth:
🤖 Follow-up on the three non-blocking items from the approval review: items 1 and 2 (doc wording, classic `DismissesStaleReviews` observation) are addressed in #461. Item 3 (`DefaultBranchRequiresPRReviews` reading only `PullRequest[0]`) is deliberately left for #459, which is already rewriting that region of `repository_metadata.go` and can delegate to the aggregating accessor in one line there.

PR #456, opened 2026-08-18
feat: publish this plugin to grc.store

eddie-knight:
Declare the publication metadata `pvtr publish` reads out of the built binary: Publisher ossf (coordinate ossf/github-repo), the Apache-2.0 license grc.store requires on every publication, and the namespace that owns each embedded catalog. The vendored OSPS Baseline YAML carries no metadata.author.id and pvtr will not guess an owner, so the mapping is declared here.

None of it is reachable at run time, so the orchestrator setup moves into newOrchestrator and a test asserts the resulting publish manifest — otherwise a wrong coordinate or a newly embedded, unmapped catalog would first surface as a failed release.

Add the publish workflow itself. It stays workflow_dispatch-only: the first successful publish TOFU-pins the signer identity to this file's path, so it must be deliberate and every later publish must run from here. PVTR_VERSION is a placeholder until the pvtr release carrying `pvtr publish` is cut.

Also carries one unrelated bugfix, kept here because it lives in the lines this PR was already moving: the `NewPluginCommands` arguments were shifted a slot, so `VersionPostfix` landed in `buildGitCommitHash` and `GitCommitHash` in `buildTime`, and `BuiltAt` — set by both `.goreleaser.yaml` and the `Makefile` — was passed nowhere. `Version` already carries the postfix, so the fix drops it and passes `GitCommitHash` and `BuiltAt` into their own slots. This changes what the plugin reports for its commit and build time.


eddie-knight (commented):
:robot: Multi-model panel review — reviewers read the PR independently in a full worktree, then every load-bearing claim was re-verified against source before it made it into this review. (Two of the three tiers reported; the third failed to return a report, so its depth was covered by direct verification instead.) Focus was the one the reviewer asked for: idiomatic, DRY, and simple comments rather than in-line documentation of everything.

**The PR does what it says.** I verified the substance rather than taking the description on trust: none of the three embedded catalogs declares `metadata.author.id`, all three `catalogNamespaces` keys match real `metadata.id` values, `PublishManifest` genuinely fails closed on an unmapped catalog (so the new test covers the stated risk), and every workflow interface — `--dist`, `PVTR_TOKEN`, `PVTR_HUB_URL`, the well-known endpoint, `ci_audience` — is real. `go vet` is clean and the test passes.

**Focus answers**

*Idiomatic* — yes. The refactor is an improvement: wrapped errors instead of `os.Exit` inside setup, and the pointer return matches `NewPluginCommands`. One stdlib swap inline (`strings.HasPrefix`), and one note about `newOrchestrator`'s `version` parameter.

*DRY* — yes, in the code. `"ossf"` five times is configuration for three distinct catalogs, not duplicated logic, and collapsing it would be worse. The repetition is in the prose: the `newOrchestrator` doc comment restates the SDK's own field docs and hardcodes a coordinate the code computes.

*Simple comments* — **this is the one to act on.** Not the volume (12–18% of added lines) but the shape: every added comment is a three-to-four-line rationale paragraph in a file whose existing style is one line — `// BuiltAt is the actual build datetime` sits four lines above a three-line block. Four of the six blocks earn their place on content (the TOFU note especially — it's load-bearing, and I confirmed it's accurate against the hub); they're just essays where the house style is a sentence. Suggestions inline. None of the comments is *wrong*, which is worth saying.

Also inline: `len(manifest.Evaluates) == 0` is weaker than the risk its own comment describes — a catalog dropping out of the embed glob wouldn't fail it.

**One thing worth fixing before merge**, inline: the publish step has no `set -euo pipefail`, so a failed token mint continues to `privateer publish` with an empty or literal-`null` token instead of aborting. Reproduced locally. Both other reviewers cleared this step as clean, which is exactly why it's here.

**Checked and NOT problems** — recorded so they don't get re-raised:
- The `curl` OIDC block does **not** duplicate anything pvtr already does — the SDK's discovery struct doesn't decode `ci_audience` and its GHA minting is hardcoded to the sigstore audience. It fills a real gap.
- `--dist dist` isn't a path mismatch (just a redundant default).
- `${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=` is correct, not a missing `?` — that URL already carries a query string.
- The `ci_audience` empty/null guard is earned: the hub emits that field as `omitempty`.
- Leading `---`, quoted `name:`, and the action SHA pins all match existing repo conventions.
- No existing behaviour damaged; `fmt.Printf`-to-stdout on error matches the deleted code.

One drive-by flagged inline: the `NewPluginCommands` args are shifted a slot and `BuiltAt` is dropped. Pre-existing and identical at merge-base — noted only because this PR moves those lines.
- Every added comment is factually accurate — I checked the load-bearing claims rather than only their length. The two-audience comment is right (`fulcioAudience = "sigstore"`, `internal/auth/signing.go:41`), and the TOFU claim matches the hub's pin behaviour.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,66 @@
...
+          if [ -z "$audience" ] || [ "$audience" = "null" ]; then
+            echo "hub ${PVTR_HUB_URL} advertises no ci_audience" >&2
+            exit 1
+          fi
+          PVTR_TOKEN=$(curl -fsSL \
+            -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${audience}" | jq -r '.value')
+          echo "::add-mask::${PVTR_TOKEN}"
+          export PVTR_TOKEN
+          privateer publish --dist dist
```
eddie-knight: No `shell:` here, so this runs under GitHub's default `bash -e {0}` — no `pipefail`. In `PVTR_TOKEN=$(curl ... | jq -r '.value')` the exit status is jq's, so a failed curl leaves `PVTR_TOKEN` empty and the step keeps going.

Two ways that lands: empty token falls through to the login store and dies with a misleading `run \`pvtr login\`` (sdk `internal/auth/login.go:71`); a 200 without a `.value` key gives the literal `null`, which passes `TrimSpace`, so pvtr sends `Authorization: Bearer null` and `::add-mask::null` masks the word "null" across the whole log.

The asymmetry is the tell — `audience` is guarded on 57-60, `PVTR_TOKEN` isn't. Every other multi-line `run:` in this repo sets the flags (`ci.yaml:32`, `sync-go-version.yaml:41`, `update-privateer-version.yml:39`).

```suggestion
        run: |
          set -euo pipefail
          audience=$(curl -fsSL "${PVTR_HUB_URL}/.well-known/grc-store-configuration" | jq -r '.ci_audience')
          if [ -z "$audience" ] || [ "$audience" = "null" ]; then
            echo "hub ${PVTR_HUB_URL} advertises no ci_audience" >&2
            exit 1
          fi
          PVTR_TOKEN=$(curl -fsSL \
            -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${audience}" | jq -r '.value')
          if [ -z "$PVTR_TOKEN" ] || [ "$PVTR_TOKEN" = "null" ]; then
            echo "failed to mint a hub OIDC token for audience ${audience}" >&2
            exit 1
          fi
          echo "::add-mask::${PVTR_TOKEN}"
          export PVTR_TOKEN
          privateer publish
```

main_test.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	if manifest.Coordinate != "ossf/github-repo" {
+		t.Errorf("coordinate = %q, want ossf/github-repo", manifest.Coordinate)
+	}
+	if manifest.License != "Apache-2.0" {
+		t.Errorf("license = %q, want Apache-2.0", manifest.License)
+	}
+	if len(manifest.Evaluates) == 0 {
+		t.Fatal("expected at least one evaluated catalog")
+	}
```
eddie-knight: `== 0` is weaker than the risk the comment above describes. An unmapped catalog is caught (`PublishManifest` errors, so `t.Fatalf` on line 18 fires), but a catalog silently dropping out of the embed glob isn't — you'd have to lose all three for this to fail. Pinning the count is what makes "a newly embedded catalog" actually noticeable, and forcing a test update when a catalog is added is the point rather than a cost.

```suggestion
	if len(manifest.Evaluates) != len(catalogNamespaces) {
		t.Fatalf("evaluated %d catalogs, want %d", len(manifest.Evaluates), len(catalogNamespaces))
	}
```
That holds while every embedded catalog needs a namespace override; if one ever ships its own `metadata.author.id`, swap it for a literal count.

main_test.go
```diff
@@ -0,0 +1,33 @@
...
+	for _, e := range manifest.Evaluates {
+		if len(e.Catalog) < len("ossf/") || e.Catalog[:len("ossf/")] != "ossf/" {
```
eddie-knight: `strings.HasPrefix` does the bounds check too.

```suggestion
		if !strings.HasPrefix(e.Catalog, "ossf/") {
```
(add `"strings"` to the import.)

main.go
```diff
@@ -33,6 +33,14 @@ var (
 		"repo",
 		"token",
 	}
+	// catalogNamespaces declares the grc.store namespace that owns each embedded
+	// catalog. The OSPS Baseline is OSSF-owned, but the vendored YAML carries no
+	// metadata.author.id and `pvtr publish` refuses to guess an owner.
```
eddie-knight: Every other comment in this `var` block is one line (`// BuiltAt is the actual build datetime`). The non-obvious fact here is just that the vendored YAML has no author id — I confirmed that's true of all three catalogs, so it's worth saying, but not in three lines.

```suggestion
	// The vendored Baseline YAML carries no metadata.author.id, so declare the owner.
```

main.go
```diff
@@ -43,10 +51,38 @@ func main() {
...
+	err = runCmd.Execute()
+	if err != nil {
+		os.Exit(shared.InternalError)
+	}
+}
+
+// newOrchestrator builds the fully populated orchestrator for the given plugin
+// version. Publisher, License and catalogNamespaces are inert at run time and
+// exist so `pvtr publish` can derive the plugin's grc.store coordinate
+// (ossf/github-repo) and its catalog linkage from the binary itself.
```
eddie-knight: This one is mostly restating things that are already written down elsewhere: "inert at run time" is lifted from the SDK's own field docs (`pluginkit/evaluation_orchestrator.go:26-39`), and the rest repeats the PR description. It also hardcodes `ossf/github-repo`, which the code derives as `Publisher + "/" + PluginName` (`publish_manifest.go:132`) — nothing keeps the prose in sync if either changes.

```suggestion
// newOrchestrator builds the orchestrator. Publisher, License and
// catalogNamespaces are unused at run time; `pvtr publish` reads them.
```

main.go
```diff
@@ -43,10 +51,38 @@ func main() {
...
+func newOrchestrator(version string) (*pluginkit.EvaluationOrchestrator, error) {
```
eddie-knight: Minor: `version` is the only injected input — `PluginName`, `RequiredVars`, `catalogNamespaces`, `files` and `dataDir` all come from package scope, and `main` assigns the `Version` global before calling anyway. The test passes `"1.2.3"` and never asserts on it (`PublishManifest` has no version field), so the parameter isn't buying testability either. Reading the global would be consistent with the rest of the function.

main.go
```diff
@@ -43,10 +51,38 @@ func main() {
...
+		os.Exit(shared.InternalError)
+	}
+
+	runCmd := command.NewPluginCommands(
+		PluginName,
+		Version,
+		VersionPostfix,
+		GitCommitHash,
+		orchestrator,
+	)
```
eddie-knight, line 64: Pre-existing, not yours — but you're moving these lines, so: the args are shifted a slot. The signature is `(pluginName, buildVersion, buildGitCommitHash, buildTime string, ...)` (`command/plugin.go:28`), so `VersionPostfix` lands in `buildGitCommitHash` and `GitCommitHash` lands in `buildTime`. `BuiltAt` is set by both `.goreleaser.yaml:28` and `Makefile:2` and never passed anywhere.

Identical at merge-base, so entirely fair to leave — flagging it while the lines are already open.

.github/workflows/publish-plugin.yaml
eddie-knight: `--dist` already defaults to `dist` (`command/harness/publish.go:58`).

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,66 @@
...
+    permissions:
+      contents: read
+      # Mints two distinct OIDC tokens: the hub push token (audience =
+      # ci_audience) and the Sigstore signing identity (audience = sigstore,
+      # requested by pvtr itself). The hub must list this repo as a
+      # ci_publishers trusted publisher for the ossf namespace.
+      id-token: write
+    env:
+      # Must be a pvtr release that contains `pvtr publish` — see REV-345.
+      PVTR_VERSION: REPLACE_WITH_PVTR_RELEASE_TAG
```
eddie-knight: Worth a literal `TODO` so it's greppable — right now the only thing marking this workflow as unrunnable is the value itself and the PR description. (Checked: the pvtr commit mounting `publish` is `24d842c`, still unreleased, so the placeholder is accurate.)

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - name: install pvtr
+        run: go install github.com/privateerproj/privateer@${PVTR_VERSION}
+      - uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
+        with:
+          version: "~> v2"
+          args: build --clean
+      - name: publish
+        # The hub token is minted and consumed inside this one step so it is
+        # never written to the job environment. pvtr reads the plugin's
+        # coordinate, license and catalog linkage from the built binary itself.
```
eddie-knight: First sentence describes what the next four lines visibly do. Second sentence is the part worth keeping.

```suggestion
        # pvtr reads the plugin's coordinate, license and catalog linkage from
        # the built binary itself.
```

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Tag or ref to build and publish (e.g. v0.28.0)"
+        required: true
+      hub-url:
+        description: "grc.store hub to publish to"
+        required: true
+        default: "https://hub.preview.grc.store"
```
jmeridth, line 20: 🤖 (medium) The free-text `hub-url` input lets the dispatcher point this job at an arbitrary hub. The script then asks that hub what audience to mint, mints a GitHub OIDC token carrying this repo's identity for that audience, and hands the token to the same hub. A hostile or typo'd URL therefore receives a genuine, replayable token, and it can advertise the real hub's audience to get one valid there. Dispatch requires write access, so the practical vector is social engineering ("run a test publish against my staging hub"), which is a much easier ask than direct publish access.

Two cheap mitigations, ideally both: make this a `type: choice` input restricted to the known preview and production URLs, and assert the fetched `ci_audience` against an expected allowlist before minting. Given the TOFU signer-pinning stakes described in the header comment, a GitHub `environment:` with required reviewers on this job would also fit.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+          persist-credentials: false
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+      - name: install pvtr
+        run: go install github.com/privateerproj/privateer@${PVTR_VERSION}
+      - uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
+        with:
+          version: "~> v2"
+          args: build --clean
```
jmeridth, line 69: 🤖 (medium) This rebuilds from source, so the bytes signed and pushed to grc.store are not the bytes the release workflow already built and attested (`release.yml` sets `create-attestation: true`). `.goreleaser.yaml:28` stamps `-X main.BuiltAt={{.Date}}` and there is no `mod_timestamp`, so a rebuild of the same tag is not byte-reproducible. Net effect: the Sigstore bundle on the hub attests a different artifact than the GitHub build provenance for the same version.

`pvtr publish --dist` supports a pre-built dist directory (the SDK's artifact loader explicitly handles a moved dist), so downloading the release artifacts for `inputs.ref` and publishing those would keep one set of bytes across both attestations.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+    permissions:
+      contents: read
+      # Mints two distinct OIDC tokens: the hub push token (audience =
+      # ci_audience) and the Sigstore signing identity (audience = sigstore,
+      # requested by pvtr itself). The hub must list this repo as a
+      # ci_publishers trusted publisher for the ossf namespace.
+      id-token: write
+    env:
+      # TODO: must be a pvtr release that contains `pvtr publish` — see REV-345.
+      PVTR_VERSION: REPLACE_WITH_PVTR_RELEASE_TAG
```
jmeridth: 🤖 (medium) Two separate things keep this workflow dead until follow-ups land, and only one is written down. The placeholder makes `go install` fail on any dispatch, which the TODO covers. But even with a real tag, no released `pvtr` has a `publish` command yet: the newest release (v0.22.0) ships benchmark/generate-plugin/install/list/run/version only, and the publish CLI exists only in SDK source. So REV-345 needs a `pvtr` release, not just a tag swap here.

Merging a guaranteed-red publish path is a legitimate choice for TOFU reasons, but it is worth making explicitly. A repo Actions variable (`vars.PVTR_VERSION`) instead of an in-file placeholder would also let the fix land without touching the TOFU-pinned workflow file.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+      ref:
+        description: "Tag or ref to build and publish (e.g. v0.28.0)"
+        required: true
+      hub-url:
+        description: "grc.store hub to publish to"
+        required: true
+        default: "https://hub.preview.grc.store"
+
+permissions:
+  contents: read
```
jmeridth: 🤖 (low, non-blocking) Deny-all at the workflow level; the job already declares `contents: read` and `id-token: write` with justifications, so this grant only widens whatever jobs get added later.

```suggestion
permissions: {}
```

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+      # ci_audience) and the Sigstore signing identity (audience = sigstore,
+      # requested by pvtr itself). The hub must list this repo as a
+      # ci_publishers trusted publisher for the ossf namespace.
+      id-token: write
+    env:
+      # TODO: must be a pvtr release that contains `pvtr publish` — see REV-345.
+      PVTR_VERSION: REPLACE_WITH_PVTR_RELEASE_TAG
+      PVTR_HUB_URL: ${{ inputs.hub-url }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
jmeridth, line 51: 🤖 (low, non-blocking) No `step-security/harden-runner` first step. Most of this repo's workflows skip it too, so this is aspirational, but this is the one job that mints OIDC tokens and pushes signed artifacts, which is exactly the egress a hardened runner exists to watch. `egress-policy: audit` costs nothing on a manual-only workflow and gives a baseline for a later `block`.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
+---
+name: "Publish plugin to grc.store"
+
+# Manual-only on purpose. The first successful publish TOFU-pins the signer
+# identity of ossf/github-repo to THIS workflow file's path, so it has to be a
+# deliberate act and every later publish must come from this same file. Add a
+# `release: {types: [published]}` trigger once the manual run is proven.
+on:
```
jmeridth, line 8: 🤖 (low, non-blocking) No `concurrency` group. Two simultaneous dispatches run two publishes in parallel, and per the header comment the first successful publish TOFU-pins the signer identity, so a concurrent-first-publish race is the worst-case version. A non-cancelling group is cheap insurance:

```yaml
concurrency:
  group: publish-plugin
  cancel-in-progress: false
```

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,70 @@
...
+        run: |
+          set -euo pipefail
+          audience=$(curl -fsSL "${PVTR_HUB_URL}/.well-known/grc-store-configuration" | jq -r '.ci_audience')
+          if [ -z "$audience" ] || [ "$audience" = "null" ]; then
+            echo "hub ${PVTR_HUB_URL} advertises no ci_audience" >&2
+            exit 1
+          fi
+          PVTR_TOKEN=$(curl -fsSL \
+            -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${audience}" | jq -r '.value')
```
jmeridth: 🤖 (low, non-blocking) The `&audience=` append is the right pattern, but the value is not URL-encoded. actions/toolkit encodeURIComponents the audience before appending; here an audience containing `&`, `#`, or whitespace would silently mint a token with a truncated `aud` (or fail the curl) rather than error clearly. `--data-urlencode` on a `-G` curl, or `jq -rn --arg a "$audience" '$a|@uri'`, closes it.

main.go
```diff
@@ -33,6 +33,12 @@ var (
 		"repo",
 		"token",
 	}
+	// The vendored Baseline YAML carries no metadata.author.id, so declare the owner.
+	catalogNamespaces = map[string]string{
+		"osps-baseline":         "ossf",
```
jmeridth: 🤖 (low, non-blocking) This key is the back-compat alias catalog (`OSPS_Baseline_backwards-compatibility-protection.yaml`, `metadata.id: osps-baseline`, same 40 requirements as 2025-10), so the publish manifest advertises the same catalog under two hub coordinates, and asserts a catalog named `ossf/osps-baseline` exists on the hub. If the alias is only meant for local config compatibility, publishing it may not be intended. Whether the hub rejects an unknown coordinate at sync time is not verifiable from here, so flagging as a design question rather than a defect.

main_test.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		t.Fatalf("PublishManifest: %v", err)
+	}
+	if manifest.Coordinate != "ossf/github-repo" {
+		t.Errorf("coordinate = %q, want ossf/github-repo", manifest.Coordinate)
+	}
+	if manifest.License != "Apache-2.0" {
+		t.Errorf("license = %q, want Apache-2.0", manifest.License)
+	}
+	if len(manifest.Evaluates) != len(catalogNamespaces) {
```
jmeridth, line 34: 🤖 (low, non-blocking) `len(manifest.Evaluates)` counts embedded catalogs, not map entries, so the two lengths are equal only coincidentally today. The unmapped-new-catalog case this test's comment describes is actually caught earlier: `PublishManifest` fails closed with "has no metadata.author.id and no CatalogNamespaces override", hitting the `t.Fatalf` above. The residual gap is the day an upstream Baseline YAML gains `metadata.author.id`: a stale or typo'd map key then silently never applies while both assertions stay green. Asserting the map keys against `evaluation_plans.SupportedCatalogIDs` (already the same three IDs) would remove that drift risk and the duplicate ID list.

jmeridth:
🤖 (summary) Multi-model review, two independent reviewers (Opus 4.8, Fable 5), every posted finding re-verified against source, SDK internals (`privateer-sdk@v1.32.2`), or a build of the PR tree before posting. Three mediums (hub-url token forwarding, rebuild-vs-attested-artifact mismatch, workflow dead until a pvtr release ships `publish`), six lows.

One thing this PR quietly fixes that deserves to be loud: `main` currently passes `VersionPostfix` into `NewPluginCommands`'s commit-hash slot and `GitCommitHash` into the build-time slot, so `version` output today reports garbage. The reordering here is verified correct against the SDK signature and by running both builds. Worth a line in the PR description.

Verified clean, for the record: all three action SHAs match their upstream version tags, no `${{ }}` interpolation inside any `run:` body, token masked before any possible output with fail-closed null checks on both curls, `set -euo pipefail` where the pipelines are, GOPATH/bin confirmed on PATH via setup-go, catalog map keys exactly match the embedded catalog IDs, the refactor is behavior-preserving on every error path, and `TestPublishManifest` runs in CI and passes on the PR tree. Also confirmed the `&audience=` URL pattern and `PVTR_HUB_URL` propagation into the SDK's viper/env lookup are both correct.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 9891a70 — `hub-url` is now a `type: choice` restricted to the preview and production hubs. The audience allowlist was skipped as redundant once the hub is pinned: both known hubs advertise `ci_audience` equal to their own `hub_url` (checked live against both `/.well-known/grc-store-configuration` endpoints). A GitHub `environment:` with required reviewers is a repo-settings change that can be layered on without touching the TOFU-pinned file, so leaving that as a follow-up.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Kept the rebuild and documented the mismatch in a comment above the goreleaser step (9891a70). Publishing the release-attested bytes is not possible with the current SDK: `pvtr publish --dist` requires a GoReleaser dist directory containing `artifacts.json` + `metadata.json` (`command/harness/publish.go:58`), which release assets do not carry. Closing the gap needs the release run's dist dir persisted (workflow artifact) or SDK support for reconstructing from release assets — deferred, but now written down where the rebuild happens.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Resolved by events: pvtr v0.22.1 (released 2026-08-19, after this review) ships `pvtr publish` — `cmd/publish.go` is in the tag, and `go install github.com/privateerproj/privateer@v0.22.1` was verified to produce a `publish` command that consumes `PVTR_TOKEN`/`PVTR_HUB_URL` exactly as this step does. 9891a70 pins `PVTR_VERSION: v0.22.1` and drops the placeholder + TODO, which also moots the `vars.PVTR_VERSION` indirection — the workflow is live on merge.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 9891a70.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 9891a70 — `harden-runner` with `egress-policy: audit` as the first step, pinned to the v2.21.0 commit.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 9891a70 — non-cancelling `publish-plugin` group.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 9891a70 — the audience is passed through `jq -rn --arg a "$audience" '$a|@uri'` before appending.

main.go
eddie-knight: :robot: Confirmed intended: the alias is a real coordinate consumers may reference, so the manifest publishes all three deliberately.

main_test.go
eddie-knight: :robot: Applied in 9891a70 — the test now asserts the `catalogNamespaces` keys against `evaluation_plans.SupportedCatalogIDs` (count + membership), so a stale or typo'd override key fails the test instead of sitting unused.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+# Manual-only on purpose. The first successful publish TOFU-pins the signer
+# identity of ossf/github-repo to THIS workflow file's path, so it has to be a
+# deliberate act and every later publish must come from this same file. Add a
+# `release: {types: [published]}` trigger once the manual run is proven.
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Tag or ref to build and publish (e.g. v0.28.0)"
```
jmeridth: :robot: (low, non-blocking) "Tag or ref" oversells what works: goreleaser without --snapshot aborts on any ref whose HEAD isn't an exact tag, and the shallow checkout fetches no tags, so a branch or SHA dispatch dies late in the run after checkout/setup/install already ran. Renaming the input to "tag" (or an early `git describe --exact-match` guard step) makes it fail fast with an honest contract. It fails closed either way — this is about the dead-run trap, not a bad publish.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,88 @@
...
+        # the built binary itself.
+        run: |
+          set -euo pipefail
+          audience=$(curl -fsSL "${PVTR_HUB_URL}/.well-known/grc-store-configuration" | jq -r '.ci_audience')
+          if [ -z "$audience" ] || [ "$audience" = "null" ]; then
+            echo "hub ${PVTR_HUB_URL} advertises no ci_audience" >&2
+            exit 1
+          fi
+          audience_uri=$(jq -rn --arg a "$audience" '$a|@uri')
+          PVTR_TOKEN=$(curl -fsSL \
```
jmeridth, line 78: :robot: (low, non-blocking) This block and the audience one above are near-identical curl-then-jq-then-null-check twins; a future edit to one (retry, error text) will predictably miss the other. Under the existing `set -euo pipefail`, `jq -er '.value // empty'` (and `.ci_audience // empty` above) fails the step on missing/null fields by itself, collapsing each five-line check to one.

main.go
jmeridth: :robot: (low, non-blocking) Mapping the compat shim here means the manifest declares both ossf/osps-baseline and ossf/osps-baseline-2025-10 as distinct Evaluates linkages for byte-identical catalog content. I get that PublishManifest fails closed on unmapped catalogs so there was no non-failing alternative, but worth confirming with the hub that the ossf/osps-baseline coordinate actually exists there before the first TOFU-pinning dispatch — otherwise that publish is the one that finds out.

main.go
```diff
@@ -43,10 +49,36 @@ func main() {
...
-		PluginName:    PluginName,
-		PluginVersion: Version,
-		PluginUri:     "https://github.com/ossf/pvtr-github-repo-scanner",
+	orchestrator, err := newOrchestrator()
+	if err != nil {
+		fmt.Printf("%v\n", err)
+		os.Exit(shared.InternalError)
+	}
+
+	runCmd := command.NewPluginCommands(
```
jmeridth, line 58: :robot: (low, non-blocking) The argument-order fix is correct against the SDK, but four consecutive same-typed string params is how VersionPostfix and GitCommitHash swapped in the first place, and nothing catches a recurrence — a future SDK bump that reorders them still compiles and TestPublishManifest stays green while released binaries report wrong provenance. A cheap guard: execute the generated version subcommand in a test and assert commit hash and build time land in the right fields. (Struct-args is the real fix, but that belongs upstream.)

jmeridth (commented):
:robot: (summary) The update lands the big ones: the pinned pvtr version, the URL-encoded audience plus the hub-url choice input, the concurrency guard on the TOFU pin, and the attestation trade-off comment. Four smaller items from my review remain — all non-blocking, inline below. Separately, DCO is failing on the merge commit (b204375, no sign-off); a re-merge or rebase with -s will clear it.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 6bf8020 — the input is now `tag` ("Tag to build and publish"), with a comment noting goreleaser refuses anything but an exact tag. Skipped the `git describe` guard step as redundant once the input stops promising refs.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 6bf8020 — both blocks collapse to `jq -er '… // empty'` under the existing `set -euo pipefail`, so a missing/null field fails the step by itself. The bespoke error messages went with the duplication; the failing command line in the step log identifies which fetch died.

main.go
eddie-knight: :robot: Checked live against both hubs (`/v1/catalogs` on hub.preview.grc.store and hub.grc.store), and you were right to flag it — none of the three declared `ossf/*` coordinates existed. The Baseline publishes as `openssf/osps-baseline` (namespace `openssf`, not `ossf`), the versioned IDs are not distinct hub coordinates (the hub versions the single catalog), and `/v1/plugins` is empty on both hubs, so no publisher namespace had precedent. Applied in b0ccce1: the plugin coordinate and all three catalog linkages now use the `openssf` namespace, matching what the hub already serves — one trusted namespace instead of two. Still prerequisite on the hub side before the first TOFU-pinning dispatch: `openssf/osps-baseline-2025-10` and `-2026-02` do not exist as coordinates yet, and this repo must be listed as a ci_publisher for `openssf`.

main.go
eddie-knight: :robot: Applied in 6bf8020 — `TestVersionProvenance` runs the generated `version` subcommand through `NewPluginCommands` with distinct sentinel strings and asserts the `Version:`, `Commit:` and `Build Time:` lines each carry the right one, so a future reorder of the four positional args fails the suite instead of shipping wrong provenance. Struct-args left for the SDK, as you said.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,83 @@
+---
+name: "Publish plugin to grc.store"
+
+# Manual-only on purpose. The first successful publish TOFU-pins the signer
+# identity of openssf/github-repo to THIS workflow file's path, so it has to be a
+# deliberate act and every later publish must come from this same file. Add a
```
jmeridth, line 6: :robot: (medium) This comment overstates the pin. grc-store-protocol's `CanonicalKeylessIdentity` strips everything from `@refs/` before pinning (confirmed against the live hub: stored `signer_identity` has no ref), so the pin is the workflow *path*, not this file's content or branch. Anyone with write access can push a branch carrying a modified `publish-plugin.yaml` and dispatch from it — same canonical identity, no review gate. A `if: github.ref == 'refs/heads/main'` condition on the job makes the implied guarantee real; the comment should say path, not file.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,83 @@
...
+    env:
+      PVTR_VERSION: v0.22.1 # first pvtr release that ships `pvtr publish`
+      PVTR_HUB_URL: ${{ inputs.hub-url }}
+    steps:
+      - uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
+        with:
+          egress-policy: audit
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.tag }}
```
jmeridth, line 53: :robot: (medium) `ref: ${{ inputs.tag }}` doesn't just build the tag, it executes it inside the job holding the OIDC request token: `.goreleaser.yaml` `before.hooks` run arbitrary shell from the checked-out tree with the full job env (reproduced locally), and `pvtr publish` execs the freshly built plugin binary with inherited env including `PVTR_TOKEN`. Since any write-access user can tag any commit without review, this is an escalation path from repo-write to the openssf publisher identity. The main-ref guard from the comment above plus requiring tags to point at main-reachable commits (or an environment with required reviewers on this workflow) would contain it.

main_test.go
```diff
@@ -0,0 +1,101 @@
...
+// NewPluginCommands takes four same-typed positional strings, so a swapped
+// argument compiles and every other test stays green while released binaries
+// report wrong provenance (it happened: VersionPostfix sat in the commit-hash
+// slot until this PR). Run the version subcommand and pin each field.
+func TestVersionProvenance(t *testing.T) {
+	orchestrator, err := newOrchestrator()
+	if err != nil {
+		t.Fatalf("newOrchestrator: %v", err)
+	}
+	cmd := command.NewPluginCommands(PluginName, "test-version", "test-commit", "test-built-at", orchestrator)
```
jmeridth, line 63: :robot: (low, non-blocking) This pins the SDK's parameter order but never exercises `main.go`'s call site — both the historical bug and a fresh globals-swap at `main.go:61-62` compile and pass the full suite (tried both). The wiring the comment says this guards is still unguarded. Extracting the call site, e.g. `newPluginCommands()` reading the package globals, and invoking that here closes it.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,83 @@
...
+        with:
+          version: "~> v2"
+          args: build --clean
+      - name: publish
+        # pvtr reads the plugin's coordinate, license and catalog linkage from
+        # the built binary itself.
+        run: |
+          set -euo pipefail
+          # jq -e fails the step on a missing/null field; pipefail covers the curls.
+          audience=$(curl -fsSL "${PVTR_HUB_URL}/.well-known/grc-store-configuration" | jq -er '.ci_audience // empty')
```
jmeridth, line 76: :robot: (low, non-blocking) My suggested collapse dropped half the old guard: `// empty` only substitutes on null/false, so `"ci_audience": ""` exits 0 and this fails much later with a misleading hub rejection instead of failing here (empirically checked; null/missing/HTML all still fail correctly). Same gap on `.value` below. Sorry about that — this restores it:
```suggestion
          audience=$(curl -fsSL "${PVTR_HUB_URL}/.well-known/grc-store-configuration" | jq -er '.ci_audience | select(. != null and . != "")')
```

main.go
```diff
@@ -43,10 +49,36 @@ func main() {
 		Version = fmt.Sprintf("%s-%s", Version, VersionPostfix)
 	}
 
-	orchestrator := pluginkit.EvaluationOrchestrator{
-		PluginName:    PluginName,
-		PluginVersion: Version,
-		PluginUri:     "https://github.com/ossf/pvtr-github-repo-scanner",
+	orchestrator, err := newOrchestrator()
+	if err != nil {
+		fmt.Printf("%v\n", err)
```
jmeridth, line 54: :robot: (low, non-blocking) The SDK's `execPublishManifest` decodes stdout and surfaces only stderr when enriching an error, so this failure path logs a bare "exit status 1" with the cause swallowed:
```suggestion
		fmt.Fprintf(os.Stderr, "%v\n", err)
```

jmeridth (commented):
:robot: (summary) Third pass, this time on the final state with the fixes themselves under scrutiny. The good news: the openssf switch, errcheck fix, and publish mechanics all check out, and curl's redirect handling and token hygiene are clean. The theme of what's left is dispatch-ref trust: the hub pins a ref-stripped workflow path (verified in grc-store-protocol source and against the live hub), and the dispatched ref controls both the workflow definition and the executed tag code, so the two medium comments below are really one trust boundary. Also, heads up: one of the round-2 fixes doesn't guard what its comment claims — details inline.

.github/workflows/publish-plugin.yaml
```diff
@@ -0,0 +1,83 @@
...
+          audience_uri=$(jq -rn --arg a "$audience" '$a|@uri')
+          PVTR_TOKEN=$(curl -fsSL \
+            -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${audience_uri}" | jq -er '.value // empty')
+          echo "::add-mask::${PVTR_TOKEN}"
+          export PVTR_TOKEN
```
jmeridth, line 82: :robot: (low, non-blocking) FYI rather than a change request: `PVTR_TOKEN` is also how the SDK's viper config resolves this plugin's required `token` var (the GitHub PAT), and `auth.BearerToken` prefers the same env over the login store. Fine inside this job, but a user who sets `PVTR_TOKEN=ghp_…` per the env-prefix convention and runs any hub command sends their PAT to the hub as a bearer. Worth an upstream SDK issue (`PVTR_HUB_TOKEN`?) before first publish makes this plugin's users the ones who find out.

main.go
```diff
@@ -33,6 +33,12 @@ var (
 		"repo",
 		"token",
 	}
+	// The vendored Baseline YAML carries no metadata.author.id, so declare the owner.
+	catalogNamespaces = map[string]string{
```
jmeridth, line 37: :robot: (low, non-blocking) Version follow-up to the coordinate thread: the manifest declares these catalogs at `2025.10`/`2026.02` (from the vendored `metadata.version`), but both live hubs only carry `v0.0.0-dev-*` releases for `openssf/osps-baseline`, and the SDK only validates non-empty. If the hub cross-checks `evaluates[].catalog_version` at sync, the first publish 422s; if it doesn't, the linkage points at versions consumers can't resolve. Worth confirming hub version semantics before the first dispatch.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 83bcad1 — the header now says the pin is the workflow *path* (the hub strips the ref), and the job carries `if: github.ref == 'refs/heads/main'`. The comment also states the guard's limit plainly: a branch copy can delete the `if:`, so containment against hostile write access remains the repo-settings environment with required reviewers, still a follow-up.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 83bcad1 — checkout now uses `fetch-depth: 0` and a guard step immediately after it refuses any tag not reachable from main, before setup-go, goreleaser's hooks, or the built binary execute anything from the tree.

The refname is fully qualified (`refs/remotes/origin/main`) deliberately: checkout's fetch-depth-0 refspec copies every upstream tag verbatim into `refs/tags/`, and a bare `origin/main` resolves through `refs/tags/` before `refs/remotes/` — so a write-access user could shadow it with a tag literally named `origin/main` and walk the guard's HEAD check straight past it (reproduced with checkout's exact refspecs; git's ambiguity warning exits 0, so `set -euo pipefail` never fires). Qualified, the same hostile clone is blocked. Also verified: the guard fails closed when origin/main is absent (exit 128), annotated tags peel HEAD to the commit, a branch name dispatched in place of a tag is still refused, and no `${{ }}` reaches any run body.

main_test.go
eddie-knight: :robot: Applied in 83bcad1 — the call site is extracted into `newPluginCommands`, which reads the package globals, and both `main` and TestVersionProvenance (sentinels set with save/restore) go through it. Verified by mutation: swapping GitCommitHash/BuiltAt, Version/GitCommitHash, PluginName/Version, or reintroducing the historical VersionPostfix-in-commit-slot each fails the test on the exact lines; restored order passes, clean under `-shuffle=on`.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Applied in 83bcad1 — both fields use `jq -er '… | select(. != null and . != "")'`. Measured exit codes: valid 0; missing key, null, empty string, empty body all 4; false 1; non-JSON 5 — every bad case aborts the step at the assignment under `set -euo pipefail`.

main.go
eddie-knight: :robot: Applied in 83bcad1 — `fmt.Fprintf(os.Stderr, "%v\n", err)`. Confirmed the mechanism against the SDK (internal/publish/helpers.go:49-58): stdout is json.Unmarshal'd and only stderr enriches the returned error, so the old Printf fed the cause into the decoder and surfaced a bare exit status.

.github/workflows/publish-plugin.yaml
eddie-knight: :robot: Checked by executing the SDK path rather than reading it, and the premise is half right. `PVTR_TOKEN` cannot resolve this plugin's `token` var at all: required vars come only from `viper.GetStringMap("vars")` merged with `targets.<name>.vars` (config/config.go:73-76), and `Config.GetString` never falls back to top-level viper keys — `PVTR_TOKEN=ghp_…` with no config file yields `missing required variables: [owner repo token]`. No env route to plugin vars exists, so the env-prefix convention can't route a PAT here.

What's real is the precedence half: `auth.BearerToken` prefers `PVTR_TOKEN` over the login store (internal/auth/login.go:70-72), and its one non-test caller is `pvtr publish` — so the hazard is a user who guessed `PVTR_TOKEN` for the PAT, hit the visible missing-vars failure, left it exported, and later ran `publish` in the same shell. That's a naming-confusion bug, not a convention trap. An upstream issue draft (rename to `PVTR_HUB_TOKEN`, deprecated `PVTR_TOKEN` fallback, matching the SIGSTORE_ID_TOKEN service-naming pattern) is ready to file against privateer-sdk.

main.go
eddie-knight: :robot: Checked against the hub source and both live hubs (preview/prod). There is no 422 path: `plugin_evaluations.catalog` is an explicitly soft, publisher-asserted reference with no FK (store/schema.sql:379-385), sync-time validation is only list-non-empty / contains-slash / requirement_ids-non-empty (internal/plugin/plugin.go:299-311), and `catalog_version` is normalized (one leading `v` stripped) and stored verbatim — never checked, not even for non-empty. The first dispatch cannot fail on this.

The unresolvable-linkage half is real and bigger than versions: `openssf/osps-baseline-2025-10` and `-2026-02` 404 on both hubs — coordinate-level, not version-level — and `openssf/osps-baseline` carries only `v0.0.0-dev-*` releases. Keeping the declared versions as-is (2025.10/2026.02): they are what the vendored `metadata.version` says, no hub read path or UI consumes `catalog_version` yet, and matching them to dev placeholders would make the manifest lie. So: hub-side prerequisite before the first dispatch, alongside the ci_publishers listing — publish real Baseline releases and decide whether the split `-2025-10`/`-2026-02` coordinates should exist as hub coordinates at all.

jmeridth (dismissed):
LGTM. conflict to fix.

jmeridth (approved):

PR #457, opened 2026-08-20
fix: install the binary in ~/.privateer/bin folder

fboudra:
Fix typo in the destination folder (missing dot). The plugin should be installed into ~/.privateer/bin folder. Make sure the destination folder is created.


eddie-knight (approved):
Thank you for catching this!

PR #459, opened 2026-08-21
feat: observe absence of branch protection without an admin token

eddie-knight:
## Summary

The `protected` boolean on the REST ["Get a branch"](https://docs.github.com/en/rest/branches/branches#get-a-branch) endpoint is readable by **any** token — verified empirically against github.com, including fully anonymous requests — and is `true` when either classic branch protection or any ruleset applies to the branch:

| Repo probed (anonymously) | Protection in place | `protected` | classic `protection` summary |
|---|---|---|---|
| ossf/scorecard | classic BP | `true` | full required-status-check list visible |
| privateerproj/homebrew-tap | deletion ruleset only | `true` | `enabled: false` |
| (unprotected fixture repo) | none | `false` | `enabled: false` |

A `false` value is therefore a **trustworthy negative** — no classic protection and no rulesets exist at all — even for a non-admin scan. A `true` value stays too weak to pass on (it may reflect only a deletion rule), so positive findings still come only from ruleset detail or admin-visible classic branch protection, as before.

## Changes

- `data`: fetch the default branch alongside the existing repository metadata (one extra concurrent REST call) and expose `DefaultBranchProtectedFlag()` plus the `ObservedUnprotected()` helper.
- `OSPS-AC`: `BranchProtectionRestrictsPushes` and `BranchProtectionPreventsDeletion` now return a confident **Failed** when the flag proves no protection exists, instead of NeedsReview asking for an admin token or Security Insights declaration.
- `OSPS-VM`: the SAST and SCA enforcement steps treat an observed absence of protection as an observable absence of required status checks, turning their "not observable with the current token" hedge into **Failed** where the absence is proven.

The residual NeedsReview cases (flag `true` but details invisible to a non-admin) are unchanged and still legitimately need an admin token or Security Insights.

## Caveats

- The docs don't spell out the field's visibility, so this is documented in-code as empirical github.com behavior; GHES may differ. A failed branch fetch degrades to the previous NeedsReview behavior.
- Follow-up candidate (not in this PR): the same endpoint exposes the classic-BP `required_status_checks` summary to non-admins, which could additionally feed the required-check matching in the QA/VM steps.

## Testing

- `go vet ./...`, `go build ./...`, `go test ./...` all pass.
- New unit cases: observed-unprotected → Failed and weak-positive → NeedsReview for both AC steps; `ObservedUnprotected` table test; branch-fetch wiring asserted in `TestLoadRepositoryMetadata`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jmeridth (commented):
:robot: (summary) The core mechanism checks out — I verified against the live API that the `protected` flag is readable by any token and that false genuinely proves absence, and permission errors can't masquerade as observed-absence. The findings are about the paths around the new guard: an admin-token fall-through to the old zero-value false Pass when the branch fetch fails, and case ordering that makes the new signal unreachable for admins. Details inline.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -83,6 +91,12 @@ func BranchProtectionPreventsDeletion(payload data.Payload) (result gemara.Resul
 		return gemara.Passed, "Default branch is protected from deletions by rulesets", gemara.High
 	}
 
+	// The publicly readable branch `protected` flag being false proves no
+	// protection of any kind exists, so deletion is not prevented.
+	if data.ObservedUnprotected(metadata) {
```
jmeridth, line 97: :robot: (high) When the new `GetBranch` call fails (403 secondary rate limit during a bulk scan, transient network error), the flag is nil, this guard doesn't fire, and an admin token falls through past the `ViewerCanAdminister` safety net to the old zero-value path: null GraphQL `RefUpdateRule` decodes `AllowsDeletions` to false and the step returns Passed/High for a repo with no protection at all. Reproduced with a unit test against this head. The fix probably wants the fetch error tracked (so nil-flag + admin + empty rules degrades to NeedsReview) rather than relying on the flag alone as the only defense against the pre-existing false Pass.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -67,6 +67,14 @@ func BranchProtectionRestrictsPushes(payload data.Payload) (result gemara.Result
 		result = gemara.Failed
 		message = "Found Ruleset, but not protection of the default branch"
 		confidence = gemara.Medium
+	// The branch `protected` flag is readable by any token, and false means no
+	// classic branch protection and no rulesets exist — a trustworthy negative
+	// even without admin. (A true flag is too weak to pass on: it may reflect
+	// only a deletion rule.)
+	case data.ObservedUnprotected(metadata):
```
jmeridth, line 71: :robot: (medium) This case is unreachable whenever the ruleset fetch succeeded: `RulesetsObserved()` is true even for a zero-length ruleset list (the `*BranchRules` is non-nil), so the case above always wins. Net effect: a non-admin scan of an unprotected repo gets this accurate message at High confidence, but an admin scan of the same repo gets "Found Ruleset, but not protection of the default branch" (no ruleset was actually found) at Medium — wrong message, inverted confidence. Moving this case above the bare `RulesetsObserved()` case fixes both.

data/repository_metadata.go
```diff
@@ -192,12 +221,19 @@ func loadRepositoryMetadata(ghClient *github.Client, owner, repo string) (ghRepo
 			branchRules = rules // hoist
 		}
 	})
+	wg.Go(func() {
+		branch, _, branchErr := ghClient.Repositories.GetBranch(context.Background(), owner, repo, repository.GetDefaultBranch(), 1)
```
jmeridth, line 257: :robot: (low, non-blocking) The "Errors are expected... safe to ignore" rationale above doesn't hold for this lookup: go-github's `GetBranch` returns an opaque `unexpected status code` error, so a rate-limit 403 is indistinguishable from a 404, and dropping `branchErr` is what opens the admin fall-through flagged in steps.go. Worth at least recording that the fetch failed so consumers can tell "observed present/absent" from "couldn't look".

evaluation_plans/osps/access_control/steps.go
jmeridth, line 71: :robot: (low, non-blocking) An archived repo with no protection now gets Failed/High "pushes are unrestricted" / "deletions are not prevented" — untrue for a read-only repo (verified against a live archived repo). `IsActive()` is on the interface but unused; gating the High-confidence phrasing on it would keep the claim factual.

jmeridth:
🤖 Heads up on interaction with #447, which touches the same two files.

**Textual**: both PRs insert into the same two anchor points of `data/repository_metadata.go` (the `RepositoryMetadata` interface block, and immediately after `RulesetsObserved()`), plus overlapping regions of `repository_metadata_test.go`. Whichever merges second gets a both-added conflict at each anchor; resolution is keep-both-sides, same shape as the #441/#445 test-file conflict.

**Semantic**: the two do not contradict, they compose. `ObservedUnprotected` is a positive, any-token observation that no protection exists, and #447's weakest branch is exactly the case it resolves: the non-admin terminal in `RequiresNonAuthorApproval` that returns `NeedsReview` with "classic branch protection is only visible to admin tokens". With this PR's signal, that branch can return a confident `Failed` when `ObservedUnprotected` is true, since no protection at all means no review requirement can exist. This PR wires the signal into access_control and vuln_management but not quality, because the QA-07 step only exists on #447's branch.

**Suggested order**: merge #447 first, then rebase this over it. The conflict would then be this PR's two blocks landing after #447's, and the rebase is a natural moment to wire the QA-07 non-admin branch to `ObservedUnprotected` since both files are already in hand. If the order goes the other way, we will rebase #447 and can adopt the signal there in the same pass. Either way, if neither PR picks it up, the QA-07 adoption deserves a small follow-up so it does not fall through the cracks.

evaluation_plans/osps/access_control/steps.go
eddie-knight: 🤖 Fixed in abce678.

The trailing `Passed` is now guarded on the `protected` flag having actually been observed. A null GraphQL `refUpdateRule` decodes to an all-zero struct that is indistinguishable from a real rule blocking deletions, so when the branch fetch failed (`DefaultBranchProtectedFlag() == nil`) and no `RefUpdateRule` field is set either, the step returns NeedsReview rather than Passed/High.

No error-tracking field was needed for this — see the reply on the `repository_metadata.go` thread.

Worth flagging that your repro was already enshrined as an expectation: the `steps_test.go` case "admin token, branch protection prevents deletion, no rulesets" asserted `Passed` for exactly this input (no `protectedFlag`, `viewerCanAdminister: true`, zero-value `AllowsDeletions`), which is why a green suite told us nothing. That case now sets `protectedFlag: &trueVal` so it tests a genuine admin-sees-protection Pass, and a new case "admin token, branch fetch failed, no rule observed" pins the NeedsReview.

evaluation_plans/osps/access_control/steps.go
eddie-knight: 🤖 Fixed in abce678 — `case data.ObservedUnprotected(metadata)` now precedes the `RulesetsObserved() && ViewerCanAdminister()` case, with a comment at the site recording why the order is load-bearing.

An admin scan of an unprotected repo now returns "Default branch has no branch protection rules or rulesets; pushes are unrestricted" at High, instead of "Found Ruleset, but not protection of the default branch" at Medium.

The pre-existing admin case at :205 did not catch this: it leaves `protectedFlag` nil, so `ObservedUnprotected` is false there regardless of ordering. Added "observed unprotected: admin token with an empty ruleset list", which sets the flag false alongside `rulesetsObserved` and `viewerCanAdminister`, so the ordering is now pinned.

data/repository_metadata.go
eddie-knight: 🤖 The distinction you asked for turns out to already exist, so abce678 keys off it rather than adding a field.

`github.Branch.Protected` is a `*bool`, and "Get a branch" always emits `protected` — the same empirical probing in the PR description saw `protected: false` come back on the unprotected fixture. So a successful fetch yields a non-nil pointer, and a nil `DefaultBranchProtectedFlag()` already means precisely "could not look". That nil is what the deletion-step guard now tests. If GitHub ever did omit the field, nil would still read as unobserved and degrade to NeedsReview, which is the safe direction.

You are right about the comment above the block, though: "Errors are expected when an org or ruleset isn't in place; safe to ignore" over-claims now that it also covers the branch goroutine, where an ignored error is exactly what we care about. Leaving that as a wording cleanup.

evaluation_plans/osps/access_control/steps.go
eddie-knight: 🤖 Confirmed and deliberately not fixed here.

`IsActive()` still has no non-test caller, so an archived repo with no protection does get Failed/High "pushes are unrestricted". Two reasons for leaving it: it is orthogonal to this PR's stated intent (that the `protected` flag is readable without admin), and it applies to the pre-existing Failed paths as much as to the new one — so a fix belongs where it can cover all of them rather than bolted onto this case.

Calling it out here so the deferral is on the record rather than silently dropped.

evaluation_plans/osps/access_control/steps.go
jmeridth: :robot: Verified fixed in abce678 — reconstructed the exact scenario (admin, empty rules, zero-valued RefUpdateRule, flag nil from a failed GetBranch) in an independent test: now NeedsReview/Low with the unobservable-protection message, no longer Passed/High. Genuine protection still passes via both the classic-rule and ruleset paths, and your regression test pins the same case. Resolving.

evaluation_plans/osps/access_control/steps.go
jmeridth: :robot: Verified fixed in abce678 — the ObservedUnprotected case now precedes the ruleset case with the load-bearing-order comment, and the admin scan of an unprotected repo returns the accurate Failed/High message instead of "Found Ruleset…" at Medium. A genuinely protecting ruleset still short-circuits to Passed first, even against a contradictory flag. Resolving.

data/repository_metadata.go
jmeridth: :robot: Verified — reusing `github.Branch.Protected` as the `*bool` distinction works: a GetBranch error leaves the flag nil, nil can never satisfy ObservedUnprotected, and both steps route nil to NeedsReview. One cosmetic remnant: the loader comment ("Errors are expected… safe to ignore") still sits above all three goroutines including the branch fetch whose error is now load-bearing as nil — worth a touch-up whenever you're next in that file. Resolving on the substance.

evaluation_plans/osps/access_control/steps.go
jmeridth: :robot: Decline accepted — verified the premise: metadata `IsActive()` still has no non-test caller (the `reusable_steps.IsActive` reference is a different function reading SI status). Archived-repo behavior stays as a known limitation. Resolving.

evaluation_plans/osps/access_control/steps.go
```diff
@@ -83,13 +92,30 @@ func BranchProtectionPreventsDeletion(payload data.Payload) (result gemara.Resul
...
 
-	if payload.Repository.DefaultBranchRef.RefUpdateRule.AllowsDeletions {
+	// An admin is not immune to the same zero-value trap: a null GraphQL
+	// refUpdateRule decodes to an all-zero struct, which is indistinguishable
+	// from a real rule that blocks deletions. The `protected` flag settles it,
+	// so when the branch fetch failed and no rule field is set either, the
+	// honest answer is that nothing was observed.
+	rule := payload.Repository.DefaultBranchRef.RefUpdateRule
+	ruleObserved := rule.AllowsDeletions || rule.AllowsForcePushes || rule.RequiredApprovingReviewCount > 0
+	if !ruleObserved && metadata.DefaultBranchProtectedFlag() == nil {
```
jmeridth: :robot: (low, non-blocking) One residual gap the fix narrows but does not close. It is pre-existing on main (reproduced there too), so not introduced here and fine as a follow-up: when the flag is observed **true** but true is explained by a non-deletion ruleset (e.g. required-status-checks only), an admin with a null RefUpdateRule still falls through to Passed/High "protected from deletions by branch protection rules" while deletions are actually unrestricted. Your own doc comment calls flag=true a weak positive, but this trailing path treats it as strong. The discriminator already exists in-data: `RulesetsObserved() && IsDefaultBranchProtectedFromDeletion() == false` with a zero-valued rule means flag=true carries no classic-BP information, and the honest answer is NeedsReview.

eddie-knight: 🤖 Fixed in b532b24 rather than deferred — the discriminator was ready-made, as you said. The guard before the trailing Passed now fires when the zero-valued rule coincides with `RulesetsObserved()`, not only when the flag is nil: at that point in the switch no observed ruleset prevents deletion, so an observed-true flag is fully explained by a non-deletion ruleset and carries no classic-BP information. New case "weak positive: admin, flag true explained by non-deletion ruleset" pins NeedsReview for your exact repro shape.

One residual, narrower still: flag=true with the ruleset fetch *failed* (rulesets unobserved) plus a zero rule still reads as the admin-sees-classic-BP Passed — the existing "admin token, branch protection prevents deletion, no rulesets" case pins that reading. Distinguishing "no rulesets" from "could not fetch rulesets" would need the ruleset error recorded, same shape as the branch-fetch discussion; flagging so it is a known gap rather than a silent one.

data/repository_metadata.go
eddie-knight: 🤖 The comment touch-up landed in b532b24 while in the file: the loader now names all three lookups and records that a branch-fetch error is kept as a nil flag rather than ignored.

eddie-knight:
🤖 #447 merged first as suggested, so this branch took it via merge (66e0751, then 8026ea2 catching up to current main), and the QA-07 adoption is done in b532b24 rather than left for a follow-up: `RequiresNonAuthorApproval` returns Failed/High on `ObservedUnprotected`, placed ahead of the ruleset-gap and admin-blind-spot terminals since the flag is decisive regardless of ruleset visibility. Pinned by "observed unprotected fails confidently for non-admin". The test fake needed a `DefaultBranchProtectedFlag()` method — its embedded nil interface would have panicked the moment the step started calling `ObservedUnprotected`.

evaluation_plans/osps/access_control/steps.go
jmeridth: :robot: Verified at b532b24: the widened guard fires for the weak-positive shape (reconstructed it empirically: NeedsReview/Low now, and reverting the guard to the nil-only form fails exactly your new pinning test, so the test is load-bearing). Genuine classic BP still passes in all three shapes I tried. Your narrower residual description is accurate: flag=true with the ruleset fetch failed plus a zero rule still reads Passed, and closing it needs ruleset fetch errors recorded, same shape as the branch-fetch work. Bonus points for finally fixing the loader comment and for the QA-07 wiring, which I also checked (positive evidence still wins; only flag-observed-false changes, from NeedsReview to a correct Failed, with a mutation-checked test). Resolving.

jmeridth (approved):

PR #462, opened 2026-08-25
docs: document AI configuration and usage

### LINKED ISSUE #319: docs: document AI configuration and usage [Draft]
## Summary

Document how to enable and use AI-assisted checks in `pvtr-github-repo-scanner`.

This issue should cover user-facing configuration, supported provider examples, guidance on validating AI configuration without provider spend, and how `[AI-Assisted]` output should be interpreted.

## Scope

- Add README documentation for AI configuration keys
- Document the supported providers (`openai` and `anthropic`)
- Explain how to validate AI configuration without incurring provider spend
- Explain timeout and token settings
- Explain how to interpret `[AI-Assisted]` results
- Update `example-config.yml` with commented AI variables

## Acceptance Criteria

- README includes an AI configuration section
- `example-config.yml` includes commented AI-related variables
- Documentation includes provider configuration examples
- Documentation explains cost and operational considerations at a high level
- Documentation explains how to validate AI configuration without provider spend
- Documentation explains `[AI-Assisted]` result labeling

## Note

This issue originally scoped "explain dry-run usage once supported by the SDK". AI dry-run was subsequently removed from `privateer-sdk`, so that bullet has been replaced. See the comment below for details.

vinayada1:
Closes #319

Documents the scanner's AI-assisted checks: why a few requirements need AI at
all, how to turn them on, how to configure a provider, and how to read the
results.

## What's here

- **`docs/ai-assisted-checks.md`** (new) — why these requirements can't be
  answered by searching, all six `ai_*` settings, where they can be placed and
  which location wins, provider examples for `openai` and `anthropic`,
  gateway/self-hosted endpoints, per-service overrides, how to check a
  configuration without paying for provider calls, how to read `[AI-Assisted]`
  results, what happens when AI cannot answer, the evidence format, size
  limits, and cost notes.
- **`README.md`** — a short section covering the settings table and result
  interpretation, linking to the deep dive.
- **`example-config.yml`** — a commented AI block split into required and
  optional settings.

Docs only; no behaviour change.

## Why AI, concretely

Rather than asserting that these requirements "can't be pattern matched", the
doc shows why, using the criteria from the prompts the scanner actually sends.
For `OSPS-QA-06.02`: documentation saying "run the tests" with no command, a
`make test` snippet that never says when tests are expected, and testing
guidance aimed at end users rather than contributors. All three contain the
keyword; none satisfies the requirement. `OSPS-QA-06.03` rests on the
difference between describing a test suite and requiring tests with changes,
and `OSPS-AC-04.02` on the fact that `contents: write` is right for a release
job and excessive for a linter.

## A note on dry-run

The issue originally asked for dry-run guidance. Dry-run does not exist in the
pinned SDK: it was added in privateer-sdk#227 and removed in privateer-sdk#252
when the AI packages were reorganised. This scanner never had a dry-run of its
own, so there was nothing to document. The issue was updated to ask instead for
guidance on validating a configuration without provider spend, and the doc has
a section on that — plus an explanation of the history so the next person
asking doesn't have to re-derive it.

## Verification

Claims were checked against the pinned `privateer-sdk` and this repo's step
implementations rather than assumed. A few were verified by running code, which
corrected several drafting errors:

- Blanking `ai_provider`/`ai_model` for one service does **not** opt it out if
  an `ai_api_key` is still reachable — the service stays enabled but invalid.
  The doc now recommends scoping AI to the services that need it.
- Misconfiguration triggers when *any* `ai_*` setting is present and a required
  one is missing, including when only an optional one such as `ai_base_url` is
  set.
- The `OSPS-AC-04.02` size limits apply only to workflows that need a judgment
  call, not to every workflow in the repository.
- Granting every scope at its highest level fails the same way as the
  `write-all` shorthand, so the summary covers both spellings.

The doc deliberately does **not** name an SDK version. A scheduled workflow
bumps that pin weekly, so a hardcoded version would go stale almost
immediately; it points at `go.mod` instead.

`go build ./...`, `go test ./...`, and `golangci-lint run` (0 issues) all pass.

Note: `docs/best-practices-badge.md` was used as the format model for the new
doc.

### DIFF
--- a/docs/ai-assisted-checks.md
+++ b/docs/ai-assisted-checks.md
@@ -10,35 +10,6 @@ optional and never replaces the deterministic checks.
 
 ## Enabling AI
 
-Add the provider and model to the service's `vars` in `config.yml`. This example
-enables OpenAI for `my-scan`:
-
-```yaml
-services:
-  my-scan:
-    plugin: github-repo
-    vars:
-      owner: <github org or user name>
-      repo: <github repo name>
-      token: <classic token with permissions repo + admin:org>
-      ai_provider: openai
-      ai_model: gpt-4o-mini
-```
-
-Set the API key in an environment variable or CI secret store, then run the
-scanner:
-
-```sh
-export PVTR_AI_API_KEY='<your-api-key>'
-./pvtr run --binaries-path .
-```
-
-All three settings are required. The scanner supports `openai` and `anthropic`.
-To use Anthropic, set `ai_provider` to `anthropic` and use an Anthropic model
-name for `ai_model`.
-
-## Configuration
-
 <!-- markdownlint-disable MD013 -->
 
 | Setting | Environment variable | Required | Default | Description |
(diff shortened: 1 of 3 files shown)


docs/ai-assisted-checks.md
```diff
@@ -0,0 +1,131 @@
...
+| `ai_timeout` | no | `30s` | Maximum time to wait for a response. |
+| `ai_max_tokens` | no | `1024` | Maximum response size. |
+
+<!-- markdownlint-enable MD013 -->
+
+Add the settings to the service's `vars` in `config.yml`. This example shows
+every setting and enables OpenAI for `my-scan`:
+
+```yaml
+services:
```
eddie-knight, line 32:
```suggestion
targets:
```

We're trying to deprecate use of `services` due to the confusion it was causing among users

docs/ai-assisted-checks.md
```diff
@@ -0,0 +1,131 @@
...
+
+Results produced with AI include the `[AI-Assisted]` prefix:
+
+```text
+[AI-Assisted] CONTRIBUTING.md explains how and when contributors should run the tests.
+```
+
+The result is `Passed`, `Failed`, or `Needs Review`, with `low`, `medium`, or
+`high` confidence. Review low-confidence results before relying on them.
+
```
jmeridth, line 96: :robot: (low, non-blocking) This path is only true for the default `output: yaml`. With `output: json` (documented as supported in example-config.yml) the file is `<service>.json`, and with `output: sarif` the AI evidence payload has no place in the SARIF schema and likely does not appear at all. Worth qualifying so users on json/sarif do not conclude the AI assessment silently failed.

jmeridth (approved):
:robot: (summary) Verified every documented config key, env var, requirement list, model default, and fallback behavior against the SDK v1.33.0 source and the scanner code, and it all checks out, including the post-#458 behavior-keyed prompts and post-#469 spellings. One small accuracy fix inline. The branch base is 9 commits behind main but the hunks land in regions #469 left untouched, so a routine rebase applies cleanly.

vinayada1:
This will be fixed as a part of PR: https://github.com/privateerproj/privateer-sdk/pull/287

PR #468, opened 2026-08-26
feat: check release tag and assets for license in LE-03.02

### LINKED ISSUE #70: Improve quality of OSPS-LE-03.02
See context in https://github.com/revanite-io/pvtr-github-repo/pull/69#issuecomment-2831059767

TLDR; the check as implemented can be more rigorous in how it finds and evaluates any license(s) associated with released artifacts

Satarupa22-SD:
fixes #70 
CC: @jmeridth


eddie-knight:
Moving to draft just while checks are failing, feel free to mark it as ready for review when the CI goes green

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
...
+	if payload.ReleasesError != nil {
+		return gemara.NeedsReview, fmt.Sprintf("Release data could not be retrieved: %v. Review the released assets for license coverage", payload.ReleasesError), gemara.Low
+	}
+
+	latest, ok := latestPublishedRelease(payload.Releases)
+	if !ok {
+		return gemara.NotApplicable, "No releases found", gemara.High
+	}
+
+	if assets := licenseAssets(latest); len(assets) > 0 {
```
jmeridth: :robot: (high) This early return makes attaching a license file to a release strictly worse than not attaching one. OSPS-LE-03.02's own text lists "a LICENSE file alongside the corresponding release assets" as a compliant way to satisfy the control (goreleaser's default template does exactly this), but any license-named asset short-circuits to NeedsReview before `GetLicenseAtRef` ever runs, so tag-level evidence that would Pass at High confidence is never consulted. And because this step is bound to both LE-03.02 and LE-02.02, gemara's aggregation drags LE-02.02 down to NeedsReview even when `GoodLicense` independently passes. Suggest checking tag-level evidence first and treating a license asset as corroborating (or at worst appending a review note), not as a blocker. Note the test `license_asset_attached_to_the_release_needs_review` currently pins the inverted behavior, so it changes too.

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
...
+		if !release.Draft {
+			return release, true
+		}
+	}
+	return data.ReleaseData{}, false
+}
+
+// licenseAssets returns the names of release assets that look like standalone
+// license files attached at the top level of the release.
+func licenseAssets(release data.ReleaseData) []string {
```
jmeridth, line 155: :robot: (medium) `licenseAssets` reuses `isLicenseFileName`, whose bare `license-` prefix rule was designed for repo-root files like LICENSE-MIT. On arbitrary asset names it flags real artifacts: `license-checker_1.0_linux_amd64.tar.gz`, `LICENSE-3rdparty.csv`, `license-report.json` all match. That contradicts the exact-match asset classifier #445 established in `build_release` (its comment says "a real artifact such as license-manager.zip is not exempted"), so the same asset gets opposite treatment in the same scan, and combined with the short-circuit above there is no path back to Passed. Reusing the #445 classifier here keeps the two consistent.

jmeridth (commented):
:robot: (summary) This is a well-built PR in the places that are hardest to get right: the data layer distinguishes 404 from error, the caching is deliberate and commented, the tests are genuinely load-bearing (we mutation-tested them), and the merge base is current. The findings are all in the verdict ladder, and the top one is the interesting kind: a behavior your own tests pin correctly but that punishes the control's recommended practice. Four models reviewed this; the inline comments are the survivors after cross-checking, several confirmed against the live GitHub API.

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
...
+	}
+
+	if latest.TagName != "" {
+		license, found, err := payload.GetLicenseAtRef(latest.TagName)
+		switch {
+		case err == nil && found && license.SpdxId != "" && license.SpdxId != noAssertion:
+			return gemara.Passed, fmt.Sprintf("GitHub identifies license %s at %q in the released source code at tag %q; the auto-generated release archives include it", license.SpdxId, license.Path, latest.TagName), gemara.High
+		case err == nil && found:
+			return gemara.Passed, fmt.Sprintf("License file %q is present in the released source code at tag %q, but GitHub could not identify the license type", license.Path, latest.TagName), gemara.Medium
+		case err == nil:
```
jmeridth, line 217: :robot: (medium) This branch maps `found == false` to Failed at High confidence, but that signal is ambiguous three ways. We confirmed against the live API that `GET /license?ref=` returns byte-identical 404s for "ref does not resolve" (deleted or re-pointed tag after publish) and "no recognized license at ref", and the endpoint only detects Licensee-conventional filenames, so a classifier miss reads the same as absence. Two of the three causes describe a correctly licensed repo. Failed/High needs the ref confirmed resolvable (a `/commits/{ref}` check) plus ideally a tag-tree filename scan analogous to `findRootLicenseFile`; otherwise NeedsReview or Medium confidence is the honest ceiling.

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
 	return gemara.Failed, "License was not found in a well known location via the GitHub API", gemara.Medium
 }
 
+// latestPublishedRelease returns the most recent non-draft release. GitHub
+// lists releases newest-first, so the first non-draft entry is the latest
+// published one. Draft releases are not published software and must not affect
+// the assessment, even when an authenticated caller can observe them.
+func latestPublishedRelease(releases []data.ReleaseData) (data.ReleaseData, bool) {
```
jmeridth, line 128: :robot: (medium) Two problems compound here: `ReleaseData` never decodes `prerelease`, so this filter cannot exclude prereleases, and `/releases` orders by `created_at`, not `published_at` (the doc comment's "newest-first" is not what GitHub does; we observed grpc return 1.83.1, 1.82.2, 1.83.0). Confirmed live on terraform: the first non-draft entry is `v1.17.0-alpha20260827` while `/releases/latest` says `v1.16.0`, so the entire verdict keys off a nightly. Adding `prerelease` to `ReleaseData` and skipping prereleases (or just using `/releases/latest`, which excludes drafts and prereleases server-side) fixes both.

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
...
+		case err == nil && found:
+			return gemara.Passed, fmt.Sprintf("License file %q is present in the released source code at tag %q, but GitHub could not identify the license type", license.Path, latest.TagName), gemara.Medium
+		case err == nil:
+			if payload.GraphqlRepoData != nil && payload.Repository.LicenseInfo.Url != "" {
+				// The tamper/removal case #70 exists to catch: HEAD is
+				// licensed, but the tree actually shipped is not.
+				return gemara.Failed, fmt.Sprintf("A license exists on the default branch, but none was found in the released source code at tag %q", latest.TagName), gemara.High
+			}
+			return gemara.Failed, fmt.Sprintf("No license was found in the released source code at tag %q", latest.TagName), gemara.High
+		default:
```
jmeridth, line 240: :robot: (medium) A failed tag lookup lands here, logs a Warn, and falls through to Passed on default-branch evidence, which the comment above rightly calls untrustworthy. GitHub signals rate-limit exhaustion with 403 (not 429), `retry.go` treats 403 as permanent, and the error is cached, so a rate-limited bulk scan disables exactly the tamper check issue #70 asks for and reports Passed. The failed `/releases` fetch above correctly yields NeedsReview; the same class of failure should get the same disposition here.

evaluation_plans/osps/legal/steps.go
```diff
@@ -117,15 +127,104 @@ func FoundLicense(payload data.Payload) (result gemara.Result, message string, c
...
+	}
+
+	if assets := licenseAssets(latest); len(assets) > 0 {
+		return gemara.NeedsReview, fmt.Sprintf("Release %q attaches standalone license file(s) as release assets (%s), which may supersede the license in the released source code; manual review is required to confirm they match an approved license", releaseLabel(latest), strings.Join(assets, ", ")), gemara.High
+	}
+
+	if latest.TagName != "" {
+		license, found, err := payload.GetLicenseAtRef(latest.TagName)
+		switch {
+		case err == nil && found && license.SpdxId != "" && license.SpdxId != noAssertion:
```
jmeridth, line 213: :robot: (low, non-blocking) The tag-level SpdxId is only used for presence; approval still reads HEAD via `GoodLicense`. So a tag whose license was swapped to SSPL-1.0 while HEAD keeps MIT passes LE-02.02 overall, with this message naming SSPL at the tag. Fine as a follow-up, but worth a note since #70 is about exactly this divergence.

data/payload.go
```diff
@@ -62,6 +62,16 @@ type payloadCache struct {
 	documentation       []DocumentationFile
 	documentationErr    error
 	documentationLoaded bool
+	// refLicenses caches LicenseAtRef lookups by ref. LE-02.02 and LE-03.02
```
jmeridth: :robot: (low, non-blocking) These comments (also line 271, and `legal/steps.go:165`) add six requirement-ID tokens to non-test code, which #448 is trying to shrink and its planned CI guard would flag. The payload layer is catalog-agnostic by design, so describing the behavior ("callers checking release licensing read the same ref twice") keeps the layer decoupled.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Verified fixed at b1c332c: tag evidence now runs first and a conclusive tag verdict decides outright; reproduced the goreleaser shape (assets plus valid MIT at tag) and got Passed/High, and reintroducing the short-circuit breaks four of your new tests. The LE-02.02 aggregation drag is gone with it. Resolving.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Verified fixed: `licenseAssets` now uses the exported #445 exact-match classifier, and `license-checker_1.0_linux_amd64.tar.gz` / `LICENSE-3rdparty.csv` no longer match (mutation-checked). The `build_release` extraction keeps the identical effective companion set, so no BR side effects. Resolving.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Verified fixed: `RefExists` disambiguates the 404 (unresolvable tag now NeedsReview/Low, mutation-checked) and genuine absence dropped to the Medium ceiling. Resolving.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Verified fixed: `prerelease` and `published_at` are decoded, prereleases are skipped, and selection is by max published time rather than list order (reproduced with an rc-first ordering; removing the filter breaks two tests). Resolving.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Partially fixed at b1c332c: the 403 path now correctly lands NeedsReview (mutation-checked), and a tag failure with a license asset present does too. But a generic failure (say a 500, which `isTransientError` also does not retry) still logs a warning and falls through to Passed/Medium on the HEAD evidence, and `steps_test.go` pins that. The inconsistency this comment flagged remains for that class: a failed `/releases` fetch gets NeedsReview while a failed tag lookup gets Passed. Is the generic-failure Passed deliberate? If not, aligning it is a one-line disposition change plus the test expectation. Minor related nit: every 403 is messaged as "rate limited", including permission-denied; the disposition is right either way, just the wording.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Confirmed still HEAD-scoped for approval; it was flagged non-blocking and works as a follow-up. Resolving.

data/payload.go
jmeridth: :robot: Verified: both payload.go comments were reworded, thanks. Two tokens remain in the `ReleasesLicensed` doc comment (steps.go:175); same one-line reword when convenient. Resolving as non-blocking.

evaluation_plans/osps/legal/steps.go
jmeridth: :robot: Verified fixed at fc98bae: a generic tag-lookup failure (500) now lands NeedsReview/Low instead of falling through to Passed on HEAD evidence, matching the failed /releases disposition. Mutation-checked: reintroducing the fall-through breaks both new regression tests. Resolving.

jmeridth (approved):
:robot: (summary) The last open thread is addressed: generic tag-lookup failures now get the same NeedsReview disposition as a failed /releases fetch, never a fallback Pass on default-branch evidence, and the new regression tests pin it (mutation-checked at fc98bae). All threads from my earlier reviews are now verified fixed or noted as follow-ups. One coordination note: #471 currently duplicates this branch's license-at-ref code; I've asked over there for it to be dropped from that PR, so expect a rebase on whichever lands second.

PR #469, opened 2026-08-28
feat: support baseline v2026.08.26

eddie-knight:
- [Applicability level format changed, so I added normalization logic](https://github.com/ossf/pvtr-github-repo-scanner/pull/469/changes#diff-6e8c4891d7d7ae9a69093667c9e3629a6425df58482816b47ae03c6fbfe0af4bR14-R32)
  - All our internal catalogs are manually overwritten with the new applicability level format (this will show up in logs also, which might be a breaking change for users who filter on log output 🤔 feedback encouraged)
  - Any previous configs that declare `Maturity Level 1` will be converted to `maturity-1` under the hood.
  - This needs to happen in the payload loader function so that it normalizes the data coming in from the user's config
- While adding the normalization logic, I figured it makes the most sense in the `catalog_contract.go` file. That file is only used within the context of `data/` so I moved it over there
- [Changed the catalog backwards compatibility protection into a floating latest](https://github.com/ossf/pvtr-github-repo-scanner/pull/469/changes#diff-4fc56285e8a5a279c91c593b6b212e79d35fd9e7a2f1a37e1a0d93a9ba9fed38)
  - [And updated the README for it](https://github.com/ossf/pvtr-github-repo-scanner/pull/469/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R7-R16)
  - We don't have any reason to continue locking in the catalog id `osps-baseline` to an old version ++ this is de-sync'd with the actual catalog releases which follow the gemara process of `PK=id,version` wherein the id does not change from `osps-baseline` in any situation. Since we are using `PK=id` we manually modify in the catalog so that `id=id+version`. Now this will have us putting the latest version into `osps-baseline` every time it changes so that the catalog's native identifier is used as a floating reference for latest.
- [Stopped using the SDK-provided logic for authorship normalization](https://github.com/ossf/pvtr-github-repo-scanner/pull/469/changes#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L36-L82).
  - I just manually copied the authorship data provided in the latest version to the older catalog yaml data, since it's functionally identical. No additional logic is needed as a safety net like we hacked in before.
- [There was a catalog change in how some assessment requirements are organized](https://github.com/ossf/pvtr-github-repo-scanner/pull/469/changes#diff-55794abd0b593b7b1405a9c6fee4de489454fa6001a9b7c7ecbf75450379a8c8L217-R236)
  - But AFAICT we already have the logic, so it's just a matter of making sure it's reflected for users of the older versions


data/catalogs/OSPS_Baseline_latest.yaml
```diff
@@ -0,0 +1,1632 @@
...
+  author:
+    id: openssf
+    name: OSPS Baseline Authors
+    type: Human
+  description: >
+    The Open Source Project Security (OSPS) Baseline is a set of security
+    criteria that projects should meet to demonstrate a strong security posture.
+  draft: true
+  gemara-version: 1.2.0
+  id: osps-baseline
```
jmeridth, line 1457: :robot: (medium) This file is a full copy of `OSPS_Baseline_2026_08.yaml` (byte-identical except this id line today), and nothing enforces that they stay equal; `catalog_contract_test.go` only checks IDs exist. That recreates the drift hazard the deleted backwards-compat file had: next bump, one file gets updated and not the other, tests stay green, and `osps-baseline` users silently evaluate a different requirement set than the versioned id while the manifest attributes both to the same upstream version. A test asserting latest matches the newest versioned catalog modulo `metadata.id` closes it cheaply.

evaluation_plans/evaluation-plans.go
```diff
@@ -39,6 +39,12 @@ var (
 		"OSPS-BR-01.01": {
 			build_release.CicdSanitizedInputParameters,
 		},
+		"OSPS-BR-01.02": {
```
jmeridth, line 42: :robot: (medium) OSPS-BR-01.02 is `state: Retired` in 2026_08 (its text is just the retirement pointer), but it is mapped to two live steps here and neither the plugin nor pluginkit filters on requirement state. A maturity-1 scan therefore runs both steps a second time, scores a retired requirement, and badge-url emits an `osps_br_01_02_status` field for something upstream deleted. Either drop the mapping for retired requirements or filter on state at dispatch.

badgeurl/badgeurl.go
```diff
@@ -223,15 +223,15 @@ func levelsForBadge(badge string) map[string]struct{} {
 	}
 
 	levels := map[string]struct{}{}
-	for _, level := range []string{"Maturity Level 1", "Maturity Level 2", "Maturity Level 3"} {
+	for _, level := range []string{"maturity-1", "maturity-2", "maturity-3"} {
```
jmeridth, line 228: :robot: (medium) `levelsForBadge` only knows `maturity-1|2|3` now, but every results file from a released build carries `Maturity Level N`, and `maturityLevelAliases` normalizes config input, not results. So `badge-url --badge baseline-1` against an existing results file filters out every log and hard-errors. Accepting both spellings here (same alias table) keeps old results files working and would also let the pinned 2025_10/2026_02 catalogs keep their original labels.

jmeridth (commented):
:robot: (summary) The #456 alias concern is handled well: verified `PublishManifest()` still emits `openssf/osps-baseline` via `metadata.author.id`, and the 2026_08 vendoring matches upstream v2026.08.28. The three inline mediums are all "what keeps this true tomorrow" problems: nothing pins `latest.yaml` to `2026_08.yaml`, a retired requirement is wired live, and pre-PR results files cannot produce badges anymore.

Smaller items, noted here rather than inline: the README catalog-contract link points at `./evaluation_plans/catalog_contract.go`, which this PR deletes (moved to `data/`). `TestPublishManifest` lost its length assertion, so a zero-Evaluates manifest now passes vacuously; asserting against `len(data.SupportedCatalogIDs)` restores it. What `osps-baseline` now resolves to is marked `draft: true` and jumps consumers from 2025.10 to v2026.08.28 content in one silent upgrade, so it deserves a release note and a conscious call on defaulting to a draft-flagged catalog. The re-vendor rewrites applicability labels inside the pinned 2025_10/2026_02 catalogs, which undercuts the pinning promise (moot if badgeurl learns the results-side alias per the inline comment). `normalizeApplicability` has no test and works partly by incidental call ordering. And the old catalogs use an `applicability-categories` key that is not a go-gemara field, so they unmarshal with zero applicability groups.

data/catalogs/OSPS_Baseline_latest.yaml
eddie-knight: This is an accepted risk. We intentionally drift from source in isolated cases to enable backwards compatibility.

evaluation_plans/evaluation-plans.go
eddie-knight: This would be a bug if it's true. We should only be running against Ids that are found in source. I will manually verify now and will update this thread with my findings.

eddie-knight: Confirmed that the code is run when using the latest catalog. Unexpected behavior from the SDK.

badgeurl/badgeurl.go
eddie-knight: applied this suggestion in a022334ad9f29eb0518784e00e74093fdd4c7aaa

evaluation_plans/evaluation-plans.go
eddie-knight: Resolved this in 540c3e7e114a302c70f77ac7c5a29660af4820cc

badgeurl/badgeurl.go
jmeridth: :robot: Verified in a022334: the alias lookup is exercised by the new test (confirmed via coverage), and the suite is green. Resolving.

evaluation_plans/evaluation-plans.go
jmeridth: :robot: I reproduced the opposite at a022334, so I want to flag this before it merges. I drove the real path (the `newOrchestrator()` construction, the real `AllSteps()` map instrumented with a counter, `Mobilize()` with catalog `osps-baseline-2026-08` and `maturity-1`, offline payload): both steps ran ("OSPS-BR-01.02 step invocations: 2") and the eval log holds a scored entry (`result=Passed stepsExecuted=2`), which aggregates into OSPS-BR-01. Nothing in privateer-sdk v1.33.0 or go-gemara v0.9.2 reads `requirement.State`; gemara filters only on the applicability intersection, and this entry carries `maturity-1/2/3`. The practical impact is bounded (both steps are already bound to BR-01.01 and BR-01.03, so this mostly double-reports), but it does produce its own logged, counted assessment with an empty recommendation. If retirement should mean not-run, the options are dropping the entry from the vendored copy, emptying its applicability, or a state filter in pluginkit. Happy to share the reproduction test if useful.

data/catalogs/OSPS_Baseline_latest.yaml
jmeridth: :robot: Agreed on intentional drift from upstream (the hand-added `metadata.author.id` is a good example). My concern was narrower: the two in-repo files (`latest` and `2026_08`) drifting from each other, since nothing asserts they stay equal. An equality-modulo-`metadata.id` test tolerates upstream drift because both copies carry it. If alias-vs-versioned divergence is itself ever intended, then accepted risk it is; otherwise the test is cheap insurance.

evaluation_plans/evaluation-plans.go
jmeridth: :robot: Our comments crossed: 540c3e7 landed just before my reply above, and it resolves this cleanly. Mapping BR-01.02 to the new `reusable_steps.Retired` step means it logs `NotRun` for visibility instead of executing and scoring live checks. Nice solution, better than the three options I listed. One nit while you are in there: the dispatch comment now says "Retired in v2026.10.25" but the vendored catalog text points at PR #443 in v2026.08.28; one of the two looks stale. The smoke-test CI failure on this head is unrelated (the Dockerfile wget of the privateer release tarball failed, exit 4). Resolving.

jmeridth (approved):

data/catalogs/OSPS_Baseline_latest.yaml
eddie-knight: Flagging this for a follow up

PR #471, opened 2026-08-31
fix: surface rest data setup errors 

### LINKED ISSUE #42: Rest data setup method not returning error conditions
Package: data
File rest-data.go

The Setup() method should take the error data from all of the rest data methods and return it up the stack for handling/logging

Satarupa22-SD:
Fixes #42

### DIFF
--- a/data/rest-data.go
+++ b/data/rest-data.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 
@@ -55,12 +56,20 @@ type RepoContent struct {
 }
 
 type ReleaseData struct {
-	Id      int            `json:"id"`
-	Name    string         `json:"name"`
-	TagName string         `json:"tag_name"`
-	URL     string         `json:"url"`
-	Draft   bool           `json:"draft"`
-	Assets  []ReleaseAsset `json:"assets"`
+	Id      int    `json:"id"`
+	Name    string `json:"name"`
+	TagName string `json:"tag_name"`
+	URL     string `json:"url"`
+	Draft   bool   `json:"draft"`
+	// Prerelease is GitHub's own flag for a release marked as not
+	// production-ready; it must be excluded from "latest published release"
+	// selection alongside drafts.
+	Prerelease bool `json:"prerelease"`
+	// PublishedAt is an RFC3339 timestamp. The /releases listing endpoint is
+	// ordered by creation time, not publish time, so selecting "the latest"
+	// release requires comparing this field rather than trusting list order.
+	PublishedAt string         `json:"published_at"`
+	Assets      []ReleaseAsset `json:"assets"`
 }
 
 type ReleaseAsset struct {
@@ -79,9 +88,11 @@ var APIBase = "https://api.github.com"
 // together (#42). Each domain degrades independently (checks fall back to
 // NeedsReview on missing data), so the caller can log the error and continue
 // the scan rather than aborting. Expected absences — the 403/404 an optional
-// feature returns when it is not available for a repository — are not failures
-// and are handled inside each fetcher.
+// feature returns when it is not available for a repository — are not
+// failures and are handled inside each fetcher.
 func (r *RestData) Setup() error {
+	// owner/repo/token are resolved by newRestData; Setup is only ever
+	// reached through it, so no fallback resolution is needed here.
 	contentsErr := r.getRepoContents()
 
 	var insightsErr, policyErr, workflowErr, vulnReportingErr, advisoriesErr error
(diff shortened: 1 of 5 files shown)


data/rest-data.go
```diff
@@ -74,35 +75,56 @@ type WorkflowPermissions struct {
...
+		vulnReportingErr = r.getPrivateVulnReporting()
 	})
 	wg.Go(func() {
-		r.getSecurityAdvisories()
+		advisoriesErr = r.getSecurityAdvisories()
 	})
 	wg.Wait()
-	return nil
+
+	return errors.Join(
```
jmeridth, line 120: :robot: (medium) The propagation this PR exists to add is not pinned by any test: I applied two mutations in a temp tree (the vuln-reporting fetcher swallowing real API errors again, and Setup discarding the `errors.Join` result and returning nil) and the full suite passed both times. The changed tests only assert `NoError` on success and expected-absence rows. One test that injects a failing fetch and asserts Setup returns a non-nil error naming that domain would lock the behavior in.

data/vuln_reporting.go
```diff
@@ -126,4 +134,23 @@ func (r *RestData) getSecurityAdvisories() {
...
+// failing). MakeApiCall formats non-200 responses as "unexpected response:
+// <code> <text>", so matching the status text inside the error message is
+// sufficient to classify the response.
+func isExpectedAbsence(err error, codes ...int) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	for _, code := range codes {
+		if strings.Contains(msg, fmt.Sprintf("%d", code)) {
```
jmeridth: :robot: (low, non-blocking) Matching `"403"`/`"404"` as substrings of the whole error text can misclassify: `MakeApiCall` and the retry wrapper embed the full endpoint URL in the message, so a repo or owner named "404" (these exist on GitHub) turns a real outage into an "expected absence" and silently drops it from the joined warn log. Matching against a typed status (or anchoring on `"status code: 404"` style substrings) avoids it. Verdicts are unaffected either way, hence non-blocking.

jmeridth (commented):
:robot: (summary) Good fix for #42, and the design nailed the part I was most worried about: the joined error is caught and Warn-logged in payload.go rather than propagating to the orchestrator, so scans still degrade gracefully instead of aborting, and every per-domain error-flag contract survives (I traced all consumers). `errors.Join` with per-domain prefixes is the right shape. Two findings inline; the first matters because it is the PR's whole purpose.

jmeridth (commented):
:robot: (summary) Both earlier findings are addressed and verified: `TestSetupPropagatesDomainErrors` pins the propagation (I re-ran the mutation check against `afbc9f2` and the test now catches it), and the typed `unexpectedStatusError` removes the substring matching, with `errors.As` surviving the retry wrapper. Nice work on both. One remaining issue, noted inline: the license-at-ref feature from #468 is duplicated into this branch and should be dropped from this PR before merge.

data/rest-data.go
```diff
@@ -523,10 +583,152 @@ func (r *RestData) getReleases() error {
 	}
 }
 
+// RefLicense describes the license GitHub detects in the repository tree at a
```
jmeridth: :robot: (medium) This commit also brings in the license-at-ref feature from #468 (`RefLicense`, `LicenseAtRef`, `RefExists`, the payload cache, the `ReleaseData.Prerelease`/`PublishedAt` fields, and their tests). That code is duplicated on the #468 branch, so whichever PR merges second will hit conflicts, and none of it is used by this fix. Can you drop the license-at-ref code from this PR and keep it in #468, where the LE-03.02 steps actually exercise it?

jmeridth: :robot: Update: #468 merged, so this branch now conflicts with main in `data/payload.go` and `data/rest-data_test.go` (the duplicated license-at-ref code and tests). Rebasing onto main and taking main's side for the duplicated hunks should resolve it, leaving just this PR's own changes: the Setup propagation test, the typed `unexpectedStatusError`, and the logging additions.

Satarupa22-SD:
Hi @jmeridth, I have fixed the merge conflicts. Please review this. Thanks!

data/vuln_reporting.go
```diff
@@ -48,45 +49,52 @@ type privateVulnReportingResponse struct {
...
 	r.PrivateVulnReporting = PrivateVulnReporting{Enabled: parsed.Enabled, Known: true}
+	return nil
 }
 
 // loadSecurityPolicy locates SECURITY.md and, when present, fetches its content.
 // Presence alone is answered from already-cached directory listings; the content
 // fetch is the only added API call, and it happens only when the file exists so
 // that its body can back the SECURITY.md contact fallback.
-func (r *RestData) loadSecurityPolicy() {
+
```
jmeridth, line 78: :robot: (low, non-blocking) The blank line here separates the doc comment from `loadSecurityPolicy`, so godoc no longer attaches it to the function.
```suggestion
```

jmeridth (approved):
:robot: (summary) All three earlier findings are addressed: the propagation test catches both regressions I tried, status checks are typed, and the license-at-ref duplication is gone after the merge. One small doc-comment nit inline. Thanks for sticking with this one.

PR #480, opened 2026-09-10
feat: add AI evaluation for SA-01.01, SA-02.01, SA-03.01, SA-03.02

vinayada1:
## Summary

Add optional AI evaluation paths for four OSPS security-assessment requirements:

| Requirement added to the AI path | Evaluation focus |
| --- | --- |
| OSPS-SA-01.01: design documentation | System actions, actors, and interactions |
| OSPS-SA-02.01: external interfaces | Documentation of external software interfaces and their contracts |
| OSPS-SA-03.01: security assessment | Identification and prioritization of likely and impactful security problems |
| OSPS-SA-03.02: threat modeling | Attack surface, critical paths, threats, and mitigations |

## Scope

**This PR's new AI paths only evaluate evidence explicitly declared in Security Insights; they do not look for additional evidence on their own.** Design review uses the declared detailed guide; interface review uses the detailed and quickstart guides; security-assessment and threat-model review use the declared self and third-party assessment evidence.

The AI evaluates the contents of those artifacts rather than treating a filename or declaration as proof of sufficient documentation. This PR does **not** add repository-wide evidence discovery, source-code scanning, arbitrary external fetching, or downstream-link traversal. Existing deterministic checks remain separate from these new AI paths.

## Evidence and result handling

- Fetch only same-repository GitHub blob/raw text URLs through the configured Contents API client, preserving the declared ref. Validate paths, formats, regular-file content, and UTF-8; do not follow redirects or download URLs.
- Cache successful and failed reads across checks. Bound retrieval to 64 KiB per document, 16 distinct URLs, and a 64 KiB assembled evidence packet.
- Defer unsupported, unavailable, comment-only, or incomplete declared evidence to NeedsReview without asking the model to grade a partial packet.
- Keep AI optional. Without AI configuration or relevant declarations, use the deterministic path. Configured-provider, configuration, and response-validation failures return NeedsReview/Low; they do not preserve an earlier Pass or Fail.
- Require affirmative documentary support for AI failures; absent or unfetched evidence alone should defer. Tighten the risk rubric to distinguish explicit likelihood-and-impact prioritization from generic security posture.
- Return Low confidence for AI NeedsReview results and cap design passes at Medium.

### Important SA-02.01 limitation

**Every SA-02.01 AI pass recommendation is returned as NeedsReview/Low, requiring human confirmation.** Live grading sometimes accepted insufficient interface documentation despite prompt tightening, so this is enforced in code, not merely requested by the prompt.

The original AI verdict, explanation, citations, and supplied material remain in the evidence for human review. AI Failed and NeedsReview responses retain their normal result handling. This intentionally also defers otherwise valid AI pass recommendations.

### Non-AI behavior changes

SA-01.01 now applies the published-release applicability gate already used by the other three checks and guards optional SI documentation fields. Its deterministic missing-design result now explicitly uses Medium confidence. Release detection still uses GitHub Releases, not tags alone or releases distributed elsewhere.

## Validation

- Passed focused security-assessment and reusable-step tests, including prompt goldens, optionality/fallbacks, and SA-02 pass deferral with Low/Medium/High model confidence while preserving original evidence.
- Passed `go test -race ./data ./evaluation_plans/osps/sec_assessment ./evaluation_plans/reusable_steps -count=1`.
- Passed targeted `golangci-lint`, `make -B build`, and `git diff --check` after the final policy change.
- Passed the native `.github/scripts/ci.sh` integration with all four supported catalog IDs.
- Full `go test ./... -count=1` and `golangci-lint run ./...` passed on the updated upstream before the final SA-02 policy change; the related checks above were rerun afterward.
- Replayed genuine SI-declared Meshery and FDC3-Sail evidence through `gpt-5.4-mini` three times each: all nine final scanner results matched the expected Meshery SA-03.01 NeedsReview, FDC3-Sail SA-01.01 Passed, and FDC3-Sail SA-02.01 NeedsReview outcomes.

The committed opt-in live test exercises production grading with caller-supplied captured fixtures; it does not replace full scanner retrieval tests or run in the normal unit suite. These finite examples do not establish general model accuracy. Existing SI reader rate limiting and legacy-schema incompatibilities remain outside this PR.


evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+				evidence.ThirdPartyAssessments = append(evidence.ThirdPartyAssessments, *declaration)
+				declarations = append(declarations, declaration)
+			}
+		}
+		for _, declaration := range declarations {
+			if declaration == nil {
+				continue
+			}
+			if declaration.Evidence == "" {
+				return evidence, nil, fmt.Errorf("security insights declares an assessment without a retrievable evidence URL")
```
jmeridth, line 703: :robot: (high) `aiAssessmentDeclaration` doesn't apply the denial filter that `assessmentDeclared` uses. With `assessments.self` containing only `comment: "No self assessment has been completed"`, the deterministic path correctly returns Failed via `commentDeniesAssessment`, but this path treats any non-empty comment as a declaration, hits this error because `Evidence` is empty, and `AIFallback` converts the earned Failed into NeedsReview/Low. Enabling AI inflates the grade for exactly the population the denial list was written to catch. `TestDeclaredEvidenceGapsPreventAIGrading` only uses affirmative comments, so the suite can't see this. Suggest skipping declarations here that `assessmentDeclared` would reject.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	}
+
+	if len(sources) > 0 {
+		aiEvidence.Description = fmt.Sprintf("AI Assisted Review of %s", strings.Join(sources, ", "))
+	}
+	payload.AddEvidence(aiEvidence)
+	if behavior == "external-interface-documentation-coverage" && response.GemaraResult() == gemara.Passed {
+		return gemara.NeedsReview, "[AI-Assisted] External interface coverage requires human confirmation; the model's pass recommendation and analysis are retained in the AI evidence", gemara.Low
+	}
+	return response.GemaraResult(), response.Summary(), securityAssessmentConfidence(behavior, response)
```
jmeridth, line 598: :robot: (high) The AI verdict returned here can overturn a deterministic verdict using a packet that never contains the artifact the verdict was based on. `declaredEvidenceURLs` only selects SI-declared guides or assessment evidence, so a repo with a thorough root `architecture.md` plus a shallow SI detailed guide can flip SA-01.01 from Passed to Failed based solely on a document the pass never depended on, while the packet asserts `collection.complete: true`. The SA-02.01 downgrade covers Pass but not Fail, so the same shape applies there. Either include the deterministically observed artifact in the packet, or only invoke AI grading from the branches whose evidence actually is the declared URL. (Related: the comment at line 64 says grading "its content" can overturn the match, but that file's content is never fetched.)

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	if clientErr == nil && client == nil {
+		return deterministic.result, deterministic.message, deterministic.confidence
+	}
+	if clientErr != nil {
+		return reusable_steps.AIFallback(payload, controlID, deterministic.message, "AI client construction failed", clientErr)
+	}
+
+	material, sources, err := loadSecurityAssessmentEvidence(payload, behavior)
+	if err != nil {
+		return reusable_steps.AIFallback(payload, controlID, "Security Insights-declared evidence could not be fully reviewed; manual review is required", "unable to gather security assessment evidence", err)
```
jmeridth, line 577: :robot: (medium) Any failure to gather declared evidence lands here and forces NeedsReview/Low, discarding a deterministic Passed. Reachable with a root `architecture.md` present: an SI detailed guide pointing at an external docs site (rejected host, probably the most common real-world value), an unsupported extension like PDF, or `payload.InsightsError` from a single YAML typo anywhere in SI (checked at line 624 before URL relevance). Each makes AI-enabled score worse than AI-disabled on the same repo, and contradicts the README's "With no relevant Security Insights declaration, it is also used unchanged." Suggest falling back to the deterministic verdict when declared evidence can't be gathered.

data/declared_documentation.go
```diff
@@ -0,0 +1,165 @@
...
+	))
+	if err != nil {
+		return DocumentationFile{}, fmt.Errorf("decode declared documentation: %w", err)
+	}
+	if len(decoded) > maxDeclaredDocumentationBytes {
+		return DocumentationFile{}, fmt.Errorf("declared documentation exceeds %d bytes", maxDeclaredDocumentationBytes)
+	}
+	text := string(decoded)
+	if !utf8.ValidString(text) || strings.ContainsFunc(text, func(r rune) bool {
+		return unicode.IsControl(r) && r != '\n' && r != '\r' && r != '\t'
```
jmeridth, line 99: :robot: (medium) `unicode.IsControl` returns false for every rune above U+00FF, so this filter admits the entire Cf category: zero-width spaces, bidi overrides, and the U+E0000 tag block used for hidden-text prompt injection. Fetched content is fully controlled by the scanned repo, and for SA-01.01/SA-03.01/SA-03.02 a model pass becomes Passed directly (at High confidence for SA-03.x, since the cap only covers design).
```suggestion
		return (unicode.IsControl(r) && r != '\n' && r != '\r' && r != '\t') || unicode.Is(unicode.Cf, r)
```
Also worth considering extending the confidence cap or pass deferral to the SA-03.x behaviors.

data/declared_documentation.go
```diff
@@ -0,0 +1,165 @@
...
+	if u.Scheme != "https" || u.User != nil || u.RawQuery != "" || u.ForceQuery || u.Opaque != "" {
+		return "", "", errors.New("declared documentation URL must use HTTPS without userinfo or query strings")
+	}
+	host := strings.ToLower(u.Host)
+	if host != "github.com" && host != "raw.githubusercontent.com" {
+		return "", "", errors.New("declared documentation URL must use github.com or raw.githubusercontent.com")
+	}
+	// Split before unescaping so a slash-containing ref stays one URL segment.
+	segments := strings.Split(strings.TrimPrefix(u.EscapedPath(), "/"), "/")
+	refIndex := 2
```
jmeridth, line 120: :robot: (medium) GitHub's Raw button now emits `https://raw.githubusercontent.com/owner/repo/refs/heads/main/docs/design.md`. With `refIndex` fixed at 2 this parses as `ref = "refs"`, `filePath = "heads/main/..."`, the fetch 404s, and the check silently lands in NeedsReview instead of being graded. Fails safe, but broken for the URL format GitHub currently hands out, and the tests only cover the legacy `/owner/repo/<ref>/path` form. Suggest consuming `refs/heads/<name>` and `refs/tags/<name>` as the ref; the Contents API accepts fully qualified refs.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	if payload.Config == nil {
+		return deterministic.result, deterministic.message, deterministic.confidence
+	}
+
+	client, clientErr := newAIClientFromConfig(*payload.Config)
+	if clientErr == nil && client == nil {
+		return deterministic.result, deterministic.message, deterministic.confidence
+	}
+	if clientErr != nil {
+		return reusable_steps.AIFallback(payload, controlID, deterministic.message, "AI client construction failed", clientErr)
```
jmeridth, line 572: :robot: (medium) This returns before `loadSecurityAssessmentEvidence` runs, so a misconfigured AI provider turns all four SA controls into NeedsReview even for repos with no relevant SI declaration, where the model would never have been consulted. The README states the opposite. Suggest selecting evidence first and returning the deterministic verdict when material is empty, treating client-construction failure as fatal only when there is something to grade.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	if err != nil {
+		return reusable_steps.AIFallback(payload, controlID, "Security Insights-declared evidence could not be fully reviewed; manual review is required", "unable to gather security assessment evidence", err)
+	}
+	if material == "" {
+		return deterministic.result, deterministic.message, deterministic.confidence
+	}
+
+	response, aiEvidence, err := reusable_steps.RunAIAssessment(client, behavior, material)
+	if err != nil {
+		return reusable_steps.AIFallback(payload, controlID, deterministic.message, "AI assessment failed", err)
```
jmeridth, line 585: :robot: (medium) `AIFallback` returns NeedsReview, but the message passed here (and at lines 572 and 588) is `deterministic.message`, so consumers can see NeedsReview paired with "Design documentation found: architecture.md" or with "...was NOT found". NeedsReview is exactly the path a human reviews, and the message misstates why it's deferred. Suggest a deferral-describing message, as the evidence-failure call at line 577 already does.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -41,7 +61,15 @@ func HasDesignDocumentation(payload data.Payload) (result gemara.Result, message
...
 					if strings.EqualFold(entry.Name, designFile) {
-						return gemara.Passed, "Design documentation found: " + entry.Name, confidence
+						// A matching filename shows a design document exists but
+						// says nothing about what it contains, so grading its
+						// content can still overturn this.
+						return gradeSecurityAssessmentDocumentation(
+							payload,
+							"OSPS-SA-01.01",
+							"design-documentation-coverage",
+							securityAssessmentVerdict{gemara.Passed, "Design documentation found: " + entry.Name, gemara.Low},
```
jmeridth, line 71: :robot: (low, non-blocking) Pre-PR, every branch of `HasDesignDocumentation` returned the unassigned named return (zero confidence value). This PR also sets Low here and on the two NeedsReview branches, beyond the stated "Medium for missing design". Arguably fixes a latent defect, just broader than the description says, and `Test_HasDesignDocumentation` still discards confidence.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +470,265 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	}
+	payload.AddEvidence(aiEvidence)
+	if behavior == "external-interface-documentation-coverage" && response.GemaraResult() == gemara.Passed {
+		return gemara.NeedsReview, "[AI-Assisted] External interface coverage requires human confirmation; the model's pass recommendation and analysis are retained in the AI evidence", gemara.Low
+	}
+	return response.GemaraResult(), response.Summary(), securityAssessmentConfidence(behavior, response)
+}
+
+// Confidence reflects the evidence, not the model's certainty about its verdict.
+// Review deferrals lack sufficient evidence, and documentary coverage claims
```
jmeridth, line 602: :robot: (low, non-blocking) This comment reads as a general rationale, but the Passed High-to-Medium cap applies only to `design-documentation-coverage`; SA-03.01/03.02 keep the model's High. The README is accurate here, the comment is what overclaims.

jmeridth:
:robot: (summary) Multi-model review (Opus 4.8, Sonnet 5, Haiku 4.5, Fable 5) at head `c45184f`, every finding verified against the source before posting.

**Verified solid:** cross-repo/off-host fetch is impossible by construction (the fetch always goes through the configured Contents API with the scanner's own owner/repo), path traversal and double-encoding are rejected, redirects are disabled on a client copy with no credential replay, size bounds hold even against a lying `size` field, the cache is per-payload with no cross-repo poisoning, the SA-02.01 pass downgrade is implemented as described, and this PR fixes a latent nil dereference in `HasDesignDocumentation`.

**Blocking theme:** verdict interaction between the deterministic and AI paths. Denial-only assessment comments regress a deterministic Failed to NeedsReview, an AI verdict can overturn a deterministic result using a packet that omits the artifact the result was based on, and ungatherable declarations (external doc URLs, PDFs, one SI YAML typo) demote deterministic Passes. Details in the inline comments: 2 high, 5 medium, 2 low (non-blocking).

Two observations not worth inline comments: refs with `%2F` can resolve fork-network commits via the Contents API (still confined to the repo's own network, so limited practical impact), and cache keys are not URL-normalized (efficiency only).

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — the assessment-declaration path now applies the denial filter (`assessmentDeclared`/`commentDeniesAssessment`), so a denial-only self-assessment no longer flips a deterministic Failed into NeedsReview; added `TestDenialCommentPreservesDeterministicFailed`.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — when declared evidence can't be gathered (rejected host, unsupported extension, or an Insights error) the check now falls back to the deterministic verdict instead of AIFallback/NeedsReview, so enabling AI never scores worse than AI-off.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — the AIFallback paths now pass deferral-describing messages instead of reusing `deterministic.message`, so NeedsReview no longer displays "Design documentation found/NOT found".

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Addressed in 794efd5 — kept the behavior (the zero-value confidence fix is intentional) and added `wantConf` confidence assertions to `Test_HasDesignDocumentation` so the Low values on those branches are now covered.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — evidence is now selected before the AI client is constructed, so a repo with no relevant SI declaration keeps its deterministic verdict even when the provider is misconfigured; client-construction failure is fatal only when there is material to grade. Added `TestMisconfiguredClientKeepsDeterministicWithoutDeclaration`.

data/declared_documentation.go
vinayada1: Fixed in 794efd5 — `parseDeclaredDocumentationURL` now consumes fully-qualified `refs/heads/<name>` and `refs/tags/<name>` refs that GitHub's Raw button emits; added raw `refs/heads` / `refs/tags` and blob `refs/heads` tests.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — an AI verdict can no longer downgrade a deterministic Passed that rests on an artifact the packet never contains (clamp; the AI analysis is still recorded as evidence). Evidence is also selected only from SI-declared URLs. Added `TestAIDoesNotDowngradeDeterministicPass`.

data/declared_documentation.go
vinayada1: Fixed in 794efd5 — the text guard now also rejects the Unicode Cf category (`unicode.Is(unicode.Cf, r)`), closing the zero-width/bidi/tag-block hidden-text injection surface; added tests for U+200B, a bidi override, and a tag-block rune.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Fixed in 794efd5 — reworded the comment to scope the High→Medium cap to design-documentation coverage; SA-03.01/03.02 keep the model's High, matching the README.

vinayada1:
All 9 findings are addressed in 794efd5 (each inline thread has a reply and is resolved); the two non-inline notes (`%2F` fork-network refs, cache-key normalization) are acknowledged as limited-impact/efficiency-only and left out of scope for this PR.

vinayada1:
Updated. Please take another look

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5. The denial filter now gates both self and third-party declarations, and TestDenialCommentPreservesDeterministicFailed covers the original scenario with zero AI calls. One small gap: only the self-assessment denial path is tested; the third-party filter shares the same predicate so I'm not blocking on it.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5 for the half that mattered: a deterministic Passed can no longer be lowered, and TestAIDoesNotDowngradeDeterministicPass covers the exact scenario. The SA-02.01 AI-Fail asymmetry remains, but since that check's deterministic verdict is never Passed and the README now documents NeedsReview-to-Failed resolution as policy, I'm treating it as accepted.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5. Both selection and gather failures preserve the deterministic triple, and the gatherFail test asserts result, message, and confidence match the AI-disabled run for all four controls.

data/declared_documentation.go
jmeridth: :robot: Verified fixed at 794efd5, with U+200B/U+202E/U+E0041 test cases. Left one follow-up comment about benign Cf characters (BOM, ZWJ) now making docs ungradable.

data/declared_documentation.go
jmeridth: :robot: Verified fixed at 794efd5 for single-segment refs on both hosts. Slash-containing branch names in fully-qualified form still misparse; left a separate non-blocking comment with the details.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5. Evidence selection is pure over Insights, runs before client construction, and TestMisconfiguredClientKeepsDeterministicWithoutDeclaration covers the original scenario for all four controls.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5. All three AIFallback sites now pass deferral-describing messages. Note the tests only assert NotEmpty on the message, so nothing pins the wording against regression; not blocking.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Confirmed at 794efd5. The wantConf assertions on all eleven cases are exactly what I was after.

evaluation_plans/osps/sec_assessment/steps.go
jmeridth: :robot: Verified fixed at 794efd5. The comment now matches the code.

README.md
```diff
@@ -52,6 +52,92 @@ docker run \
...
+fragments select no smaller scope: the entire declared file is reviewed.
+Supported text formats are Markdown, AsciiDoc, reStructuredText,
+plain text, JSON, YAML, and Protocol Buffers. Links within an artifact are not
+followed.
+
+- Without AI configuration, no additional evidence is fetched and deterministic
+  evaluation is used. With no relevant Security Insights declaration, it is also
+  used unchanged; the scanner does not search for alternative AI evidence, and a
+  misconfigured provider cannot change the verdict for such a repository.
+- Enabling AI never produces a worse verdict than the AI-disabled path. When
```
jmeridth, line 84: :robot: (medium) This sentence overstates the guarantee. With declared evidence present, client-construction failure, provider failure, and schema-validation failure (steps.go 585/607/610) all return NeedsReview/Low even when the deterministic verdict was Passed, and the "configuration" case in TestSecurityAssessmentAIFailuresNeedReview locks that in for the design control. It also inverts incentives: a provider outage now scores worse than an actual AI fail, which the Passed guard blocks. Either scope this sentence to evidence-gathering failures (the bullet below already documents the deferrals), or extend the Passed guard to those three fallback paths.

data/declared_documentation.go
```diff
@@ -0,0 +1,180 @@
...
+		return "", "", errors.New("declared documentation URL must reference the assessed repository")
+	}
+	if host == "github.com" && segments[2] != "blob" {
+		return "", "", errors.New("declared documentation GitHub URL must use /blob/<ref>/<path>")
+	}
+	// GitHub's "Raw" button and blob URLs can carry a fully-qualified ref such as
+	// refs/heads/<branch> or refs/tags/<tag>; consume all three segments as the
+	// ref (the Contents API accepts fully-qualified refs) so the file path starts
+	// after it rather than being mis-split into "heads/<branch>/...".
+	refEnd := refIndex + 1
```
jmeridth, line 147: :robot: (low, non-blocking) Two edge cases in the fully-qualified form: a slash-containing branch misparses (`refs/heads/release/v2/CHANGELOG.md` becomes ref `refs/heads/release`, path `v2/CHANGELOG.md`, normally a 404 that silently skips AI review, and if a `release` branch happens to contain `v2/CHANGELOG.md` the scanner grades a different document than the URL denotes), and the README's `%2F` workaround is rejected here because only the segment at refIndex may contain a decoded slash. `refs%2Fheads%2Ffeature%2Fx` as one segment works but isn't documented. Suggest permitting `%2F` in the segment after `refs/heads`/`refs/tags`, or documenting the limitation.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +471,318 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+) (gemara.Result, string, gemara.ConfidenceLevel) {
+	if payload.Config == nil {
+		return deterministic.result, deterministic.message, deterministic.confidence
+	}
+
+	// Decide whether there is declared evidence to grade without fetching, so an
+	// AI-disabled repo never triggers network access and a repo with no relevant
+	// declaration keeps its deterministic verdict even if the provider is
+	// misconfigured.
+	urls, selErr := declaredSecurityAssessmentURLs(payload, behavior)
```
jmeridth, line 576: :robot: (low, non-blocking) `selErr` is swallowed silently while the gather path at 596 logs a Warn. This also hides `unknown security assessment behavior`, so a future typo in a behavior name would silently disable AI review for that control forever. Suggest logging `selErr` when non-nil.

data/declared_documentation.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	if len(decoded) > maxDeclaredDocumentationBytes {
+		return DocumentationFile{}, fmt.Errorf("declared documentation exceeds %d bytes", maxDeclaredDocumentationBytes)
+	}
+	text := string(decoded)
+	if !utf8.ValidString(text) || strings.ContainsFunc(text, func(r rune) bool {
+		// unicode.IsControl only covers the Cc category, so also reject the Cf
+		// (format) category: zero-width spaces, bidi overrides, and the tag block
+		// used to smuggle hidden-text prompt injection past a human reviewer.
+		return (unicode.IsControl(r) && r != '\n' && r != '\r' && r != '\t') || unicode.Is(unicode.Cf, r)
```
jmeridth, line 102: :robot: (low, non-blocking) The Cf rejection also catches benign content: U+FEFF (a doc saved with a UTF-8 BOM), U+200D (every compound emoji), and soft hyphens now make a document ungradable (it falls back to the deterministic verdict, so no damage, just silent skips). Suggest stripping a leading BOM before validation. Also worth knowing the hidden-text channel isn't fully closed (variation selectors are Mn, Hangul fillers are Lo, U+2800 is So), so this guard is defense-in-depth on top of the prompt hardening, as the comment above implies.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +471,318 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	}
+	declarations, urls, err := declaredEvidenceURLs(payload.Insights, behavior)
+	if err != nil {
+		return "", nil, err
+	}
+	if len(urls) == 0 {
+		return "", nil, nil
+	}
+	packet := securityAssessmentAIEvidence{
+		SecurityInsights: declarations,
```
jmeridth, line 679: :robot: (low, non-blocking) The SI `name`/`comment` strings enter the prompt packet without the UTF-8/control/Cf filter applied to fetched documents, so injection text in `security-insights.yml` bypasses that guard entirely. Prompt hardening still applies; suggest running these through the same text filter for consistency.

evaluation_plans/osps/sec_assessment/steps.go
```diff
@@ -344,15 +471,318 @@ func HasThreatModelAnalysis(payload data.Payload) (result gemara.Result, message
...
+	}
+	if err := reusable_steps.ValidateAIResponse(response); err != nil {
+		return reusable_steps.AIFallback(payload, controlID, "AI-assisted review returned a response that did not conform to the expected verdict schema; manual review is required", "AI response did not conform to the expected verdict schema", err)
+	}
+
+	if len(sources) > 0 {
+		aiEvidence.Description = fmt.Sprintf("AI Assisted Review of %s", strings.Join(sources, ", "))
+	}
+	payload.AddEvidence(aiEvidence)
+	if behavior == "external-interface-documentation-coverage" && response.GemaraResult() == gemara.Passed {
```
jmeridth, line 617: :robot: (low, non-blocking) This branch runs before the deterministic-Passed guard at 622. Harmless today because SA-02.01's deterministic verdict is never Passed, but if it ever gains one, an AI pass would downgrade it while an AI fail would be blocked. Moving the guard above this branch removes the latent inversion.

jmeridth:
:robot: (summary) Re-review of `794efd5` (Opus 4.8 + Fable 5, every claim verified against the code and tests before posting).

All nine prior findings are fixed, each with a test covering the original failure scenario; both highs are solidly closed. I replied on each thread with the verification details.

New in this pass: one medium (README line 84 promises more than the three fatal-deferral paths deliver) and five low non-blocking notes, all as line comments.

Two things worth a pass outside the diff: the PR description is stale relative to this commit (three bullets describe removed behavior, and the new "AI cannot lower a deterministic Passed" rule, the headline of the fix, isn't mentioned), and the rationale at README lines 96-99 doesn't hold when the DetailedGuide points at the root document itself (the packet then does contain the artifact the Passed rests on; the policy outcome is still documented, so this is a wording fix).

With the README sentence scoped and the description refreshed, this is in good shape.

README.md
vinayada1: Accepted, and I took the stronger option you offered: the deterministic-Passed guard now covers the client-construction, provider, and schema-validation deferrals rather than scoping the sentence down. Fixed in 071af37 via a centralized `securityAssessmentResultWithPassFloor`, the `TestSecurityAssessmentAIFailuresNeedReview` configuration case was updated, and the README bullets were corrected to match.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Accepted. In 071af37 the pass floor is applied after the SA-02.01 pass cap rather than before it, so the external-interface branch can no longer invert the guard if that check ever gains a deterministic Passed, and AI evidence is still recorded when the grade is clamped.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Accepted. `selErr` is now logged with a Warn before the deterministic verdict is returned, mirroring the gather path, so an `unknown security assessment behavior` typo is visible instead of silently disabling AI review for that control.

data/declared_documentation.go
vinayada1: Partially accepted. I took the explicit suggestion and now strip a single leading UTF-8 BOM before validation in 071af37, so a BOM-saved document is gradable and the BOM is absent from the returned content. I deliberately kept rejecting the rest of Cf, including ZWJ and soft hyphens: these documents fail safe to the deterministic verdict, and you are right that this guard is defense-in-depth on top of the prompt hardening rather than a complete hidden-text channel, so I would rather keep it strict than widen it.

evaluation_plans/osps/sec_assessment/steps.go
vinayada1: Accepted. In 071af37 the SI `name` and `comment` strings are run through a sanitizer that drops control and Cf runes before they enter the prompt packet, so `security-insights.yml` content is filtered consistently with fetched documents. It strips rather than rejects, since dropping the whole declaration would change grading, and it is a no-op on clean input so the golden prompt fixtures are unchanged.

data/declared_documentation.go
vinayada1: Accepted the code fix rather than only documenting it. In 071af37 a percent-encoded slash is permitted in the branch or tag name segment following `refs/heads` or `refs/tags`, so `refs/heads/release%2Fv2/CHANGELOG.md` resolves to ref `refs/heads/release/v2` and path `CHANGELOG.md`; file path segments still reject encoded slashes. The unencoded `refs/heads/release/v2/...` form stays as-is because it is genuinely ambiguous and fails safe, and the README now documents the encoding rule.

data/declared_documentation.go
```diff
@@ -0,0 +1,182 @@
...
+	default:
+		if !isDocumentationPath(filePath) {
+			return "", "", errors.New("declared documentation has an unsupported file extension")
+		}
+	}
+	return ref, filePath, nil
+}
+
+func validDeclaredDocumentationSegment(segment string, allowSlash bool) bool {
+	if !utf8.ValidString(segment) || strings.ContainsAny(segment, "\\%") || strings.ContainsFunc(segment, unicode.IsControl) {
```
jmeridth, line 170: :robot: (low, non-blocking) Filenames and refs in declared URLs can still carry Cf runes, so hidden text reaches the prompt through `path` and `source_url` and skips the filter the document body and SI strings now get. `unicode.IsControl` only matches Cc, so a zero-width or tag-block character in a committed filename passes this check. Would rejecting Cf here too, like the content check does, close that gap?

jmeridth (approved):
:robot: (summary) All six round-2 findings are addressed in 071af37, and extending the pass floor to the provider, config, and schema failures was the stronger fix. One small note inline about Cf runes in URL paths. Before merge, could you refresh the description? Two bullets still describe the old behavior: unavailable evidence now keeps the deterministic verdict, and provider/config/validation failures now preserve a deterministic Passed.

